Issue #145, opened 2019-04-04
dev_atom is not getting created upon composing up docker instances in Macbook


dhruvkakadiya:
I am using Macbook with High Sierra OS and I tried to follow steps given in the README file but they are not exactly working for me.

Upto *Building Containers*, everything is working fine but after executing `docker-compose up -d` command, it's creating `atom_atom_1` and `atom_nucleus_1` instances. So I can't run `docker exec -it dev_atom bash` command successfully as there is no `dev_atom` named container.

```
▶ docker exec -it dev_atom bash
Error: No such container: dev_atom
▶ docker container ls
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                              NAMES
fe7791a53cfd        atom_atom           "tail -f /dev/null"      55 seconds ago      Up 53 seconds       5900/tcp, 0.0.0.0:6080->6080/tcp   atom_atom_1
25c01d564176        atom_nucleus        "docker-entrypoint.s…"   56 seconds ago      Up 54 seconds       6379/tcp                           atom_nucleus_1

▶ docker exec -it atom_atom_1 bash
root@fe7791a53cfd:/atom# atom-cli
    ___  __________  __  ___   ____  _____
   /   |/_  __/ __ \/  |/  /  / __ \/ ___/
  / /| | / / / / / / /|_/ /  / / / /\__ \ 
 / ___ |/ / / /_/ / /  / /  / /_/ /___/ / 
/_/  |_/_/  \____/_/  /_/   \____//____/  
                                          

                                                                                                                                                                                  
> 
```

So either documentation needs to be updated(if same case is applicable for Ubuntu/Windows OS) or docker script needs to be changed such that it doesn't use some name by default but it stays the same across different OS.

dpipemazo:
Thanks for the detailed bug report! You're correct, there's a bug in the documentation, the correct command should be:
```
docker-compose -p dev up -d
```

The `-p dev` part sets the "project name" of the docker-compose session. If the flag isn't passed, it auto-generates the project name from the current folder (atom, in this case). This will almost always be atom, but you'll notice that if you rename the folder in which you cloned the repo your containers would have names that look like `$my_folder_name_atom_1`.

I'll update the docs! Thanks again for the report.

dpipemazo:
[...]

PR #146, opened 2019-04-05
README: fix bug in docs

dpipemazo:
Project name wasn't being specified. Addresses bug from #145


dpipemazo:
Great, thanks for the catch on that. We tried protobuf originally and then removed support; I'll file a new ticket to clean up the docs for that.

PR #151, opened 2019-04-12
Append UUID4 hex string to atom-cli element name

cheripai:
This assures that every atom-cli instance has a unique element name, even when multiple instances are launched on the same device/container.


cheripai:
UUID is a standard python library

PR #153, opened 2019-05-24
152/fixing cmd res issue

MarquiseRosier:
Added Try-Except block in format_records function
Added Try_Except blocks in mode_cmdres to handle possible exceptions from Decoding Data. 
Also made deserialization dependent on actual flag set by user of atom-cli
And added measure to turn it off if deserializing entries proves fallable. 

Next steps, create an issue for the TypeError that occurs sometimes. with error message: 
<class 'TypeError'> Object of type 'bytes' is not JSON serializable

This should be a relatively quick fix!!


cheripai:
Thanks for getting this PR in :smiley_cat: 
Seems like we're properly displaying responses whether the data is msgpack'd or not with the `records` command, which is awesome.
However, it seems like msgpack'd command data is not displayed properly, so let's get that working before we get this PR merged in

PR #156, opened 2019-06-04
Fix for concurrency bug in python library, and corresponding test

plipay:
Discovered a bug in the Element class, where we get redis client crashes if we try to publish at >5-10 hz to a stream on one thread, while running the command_loop and handling incoming commands on another thread. This is a paradigm we rely on in some of our other element repos, (for example: https://github.com/elementary-robotics/element-instance-segmentation/blob/master/run.py), but up till now we haven't been running them fast enough to hit this consistently. After some digging, I traced the problem to the use of a shared redis `_pipeline` inside the Element class. From the thread safety section of the pyredis docs: `It is not safe to pass PubSub or Pipeline objects between threads`: (https://github.com/andymccurdy/redis-py/blob/master/README.rst#thread-safety)

This PR addresses the issue by having the write functions use their own dedicated pipeline object instead of the one used for the command loop/reader calls. This seems like a good compromise for now, where we aren't slowing down the write/read calls with the use of mutexes, and also aren't requiring any api changes from any elements that depend on atom. I've included a unit test which validates this error case, you should see it fail when running the old version.

I've also modified the command_loop while condition to make it possible to trigger a shutdown using an Event, mostly as a convenience for the unit test, but also it doesn't hurt to have this capability.

### DIFF
--- a/languages/python/atom/element.py
+++ b/languages/python/atom/element.py
@@ -1,4 +1,5 @@
 import redis
+import threading
 from atom.config import DEFAULT_REDIS_PORT, DEFAULT_REDIS_SOCKET
 from atom.config import LANG, VERSION, ACK_TIMEOUT, RESPONSE_TIMEOUT, STREAM_LEN, MAX_BLOCK
 from atom.config import ATOM_NO_ERROR, ATOM_COMMAND_NO_ACK, ATOM_COMMAND_NO_RESPONSE
@@ -25,12 +26,14 @@ class Element:
         self.timeouts = {}
         self.streams = set()
         self._rclient = None
+        self._command_loop_shutdown = threading.Event()
         try:
             if host is not None:
                 self._rclient = redis.StrictRedis(host=host, port=port)
             else:
                 self._rclient = redis.StrictRedis(unix_socket_path=socket_path)
             self._pipe = self._rclient.pipeline()
+            self._write_pipe = self._rclient.pipeline()
             self._pipe.xadd(
                 self._make_response_id(self.name),
                 maxlen=STREAM_LEN,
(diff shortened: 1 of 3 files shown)


languages/python/atom/element.py
```diff
@@ -25,12 +26,14 @@ def __init__(self, name, host=None, port=DEFAULT_REDIS_PORT, socket_path=DEFAULT
         self.timeouts = {}
         self.streams = set()
         self._rclient = None
+        self._command_loop_shutdown = threading.Event()
```
dpipemazo: :+1: nice

languages/python/atom/element.py
```diff
@@ -25,12 +26,14 @@ def __init__(self, name, host=None, port=DEFAULT_REDIS_PORT, socket_path=DEFAULT
...
         try:
             if host is not None:
                 self._rclient = redis.StrictRedis(host=host, port=port)
             else:
                 self._rclient = redis.StrictRedis(unix_socket_path=socket_path)
             self._pipe = self._rclient.pipeline()
+            self._write_pipe = self._rclient.pipeline()
```
dpipemazo: :+1: nice. This is a good catch. This should help us multithread for one reader and one writer, though I think it still might not work with multiple readers? On the C++ API I made a pool of 20 or so pipes that I give to whichever thread is requesting. Not sure we have/want to do this here, but we should perhaps understand and well document the thread-safety of the class now that we have a better idea.

languages/python/tests/test_atom.py
```diff
@@ -213,7 +213,50 @@ def test_read_since(self, caller, responder):
         entries = caller.entry_read_since("test_responder", "test_stream", last_id, 2)
         assert(len(entries) == 2)
         assert entries[-1]["data"] == b"1"
-        
+
+    def test_parallel_read_write(self, caller, responder):
```
dpipemazo: :+1: nice test

languages/python/atom/element.py
plipay: The pool of pipelines is definitely a good idea. You are right that the current implementation can still break down with multiple readers/writers. I'll look at that as a follow-on task and make an issue for it.

PR #160, opened 2019-06-05
159/msg pack false missing

MarquiseRosier:
Image of problem being fixed has been attached here!

SnapshotOfWorkingCode


utilities/atom-cli/atom-cli.py
```diff
@@ -245,12 +258,19 @@ def mode_cmdres(self, start_time, elements):
...
+            cur_records = self.element.entry_read_since(
+                None, stream, start_time, deserialize=False)
             for record in cur_records:
-                record = { key:(value if isinstance(value,str) else value.decode()) for key,value in record.items() } # Decode strings only which are required to
+                for key, value in record.items():
+                    try:
+                        if not isinstance(value, str):
+                            value = value.decode()
+                    except:
+                        value = msgpack.unpackb(value)
```
plipay: Is this change necessary? Just making sure this is intentional as behavior is different than the old record assignment.

Issue #46, opened 2018-12-05
Init heartbeat + dependencies


dpipemazo:
In all language clients, spin up a thread that publishes a heartbeat at 1Hz in the init function for the element. Also change the init function to take a list of dependencies and then wait for heartbeats of all dependencies before publishing one's own heartbeat.

PR #162, opened 2019-06-07
46/Python support for healthchecks and wait_for_elements_healthy

plipay:
https://github.com/elementary-robotics/atom/issues/46

Implements support for atom elements to have custom healthchecks, which is done using the existing command_loop. This change should be backwards compatible as long as existing users don't name any of their commands `healthcheck`. If they do happen to do this, they'll get a descriptive error message when they upgrade. Elements default to reporting healthy as soon as their command handler loop is launched.

This PR also adds support for waiting for a list of elements you depend on to be online and reporting healthy, with the new `wait_for_elements_healthy` function. 

I've added tests for these new features and updated the atombot walkthrough to demonstrate how to use it. Let me know if you see any red flags.


doc/files/walkthrough/atombot.py
```diff
@@ -111,6 +116,10 @@ def get_pos_map(self):
     # Instantiate our AtomBot class
     atombot = AtomBot()
 
+    # We add a healthcheck to our atombot element.
+    # This is optional. If you don't do this, atombot is assumed healthy as soon as its command_loop executes
+    element.healthcheck_set(atombot.is_healthy)
```
dpipemazo: :+1: this is awesome! Clever idea to be able to write your own healthcheck.

languages/python/atom/element.py
```diff
@@ -54,6 +56,10 @@ def __init__(self, name, host=None, port=DEFAULT_REDIS_PORT, socket_path=DEFAULT
             # Keep track of command_last_id to know last time the element's command stream was read from
             self.command_last_id = self._pipe.execute()[-1].decode()
 
+            # Init a default healthcheck, overridable
+            # By default, if no healthcheck is set, we assume everything is ok and return error code 0
+            self.healthcheck_set(lambda: Response())
```
dpipemazo: :+1: Nice

languages/python/atom/element.py
```diff
@@ -200,9 +206,48 @@ def command_add(self, name, handler, timeout=RESPONSE_TIMEOUT, deserialize=False
         """
         if not callable(handler):
             raise TypeError("Passed in handler is not a function!")
+        if name == HEALTHCHECK_COMMAND:
```
dpipemazo: Nice check

languages/python/atom/element.py
```diff
@@ -200,9 +206,48 @@ def command_add(self, name, handler, timeout=RESPONSE_TIMEOUT, deserialize=False
...
 
+    def healthcheck_set(self, handler):
+        """
+        Sets a custom healthcheck callback
+
+        Args:
+            handler (callable): Function to call when evaluating whether this element is healthy or not.
+                                Should return a Response with err_code ATOM_NO_ERROR if healthy.
+        """
+        if not callable(handler):
```
dpipemazo: :+1: nice check

languages/python/atom/element.py
```diff
@@ -200,9 +206,48 @@ def command_add(self, name, handler, timeout=RESPONSE_TIMEOUT, deserialize=False
...
+        if name == HEALTHCHECK_COMMAND:
+            raise ValueError(f"'{HEALTHCHECK_COMMAND}' is a reserved command name dedicated to healthchecks, choose another name")
         self.handler_map[name] = {"handler": handler, "deserialize": deserialize}
         self.timeouts[name] = timeout
 
+    def healthcheck_set(self, handler):
+        """
+        Sets a custom healthcheck callback
+
+        Args:
```
dpipemazo: great docs

languages/python/atom/element.py
```diff
@@ -200,9 +206,48 @@ def command_add(self, name, handler, timeout=RESPONSE_TIMEOUT, deserialize=False
...
+            handler (callable): Function to call when evaluating whether this element is healthy or not.
+                                Should return a Response with err_code ATOM_NO_ERROR if healthy.
+        """
+        if not callable(handler):
+            raise TypeError("Passed in handler is not a function!")
+        # Handler must return response with 0 error_code to pass healthcheck
+        self.handler_map[HEALTHCHECK_COMMAND] = {"handler": handler, "deserialize": False}
+        self.timeouts[HEALTHCHECK_COMMAND] = RESPONSE_TIMEOUT
+
+    def wait_for_elements_healthy(self, element_list, retry_interval=HEALTHCHECK_RETRY_INTERVAL):
```
dpipemazo: One idea here: We have an atom "version" and "language" that each element puts onto their request/response queue at startup. We could use this to gradually roll out the healthcheck s.t. we can know if an element fails a healthcheck because it doesn't support it or something is actually wrong. Otherwise, at the moment, for users to use this they need to know which language client was used to implement the element.

Perhaps we add a "version" command by default as well? We call the version command and if this fails we know that the element doesn't support the versioning system and therefore doesn't support any version-specific features. Otherwise the element responds with:
```
{ "language": "X", "version": Y }
```
and we can use that logic to determine if a healthcheck is supported? 

This might be a bit complex. Alternatively we could have an element in the nucleus that tracks which features are implemented in which versions of language clients, s.t. we can make an element that can be called with "supports" with a call such as:
```
element.command_send("features", "supports", data={"element": "robot", "feature": "healthcheck"})
```
which would then let the caller know if the robot supports the healthcheck? This element would then keep track of the table of (language client, version) to know when any feature was implemented so it's kept globally in one place.

Likely a bit much and went down a bit of a rabbit hole here, but interested to hear your thoughts. For now it seems we can on the user side restrict calling this function for elements that have been recently rebuilt with the python client.

languages/python/atom/element.py
```diff
@@ -248,9 +293,14 @@ def command_loop(self):
                     err_code=ATOM_COMMAND_UNSUPPORTED, err_str="Unsupported command.")
             else:
                 try:
-                    data = unpackb(
-                        data, raw=False) if self.handler_map[cmd_name]["deserialize"] else data
-                    response = self.handler_map[cmd_name]["handler"](data)
+                    if cmd_name != HEALTHCHECK_COMMAND:
```
dpipemazo: does this need to be broken out? Seems if we note there's no serialization on the healthcheck we might be able to get rid of this branch

languages/python/tests/test_atom.py
```diff
@@ -257,6 +258,83 @@ def entry_write_loop(responder):
             responder_0._rclient.delete("command:responder_0")
             responder_0._rclient.delete("response:responder_0")
 
+    def test_healthcheck_default(self, caller, responder):
```
dpipemazo: :+1: really solid tests

languages/python/atom/element.py
plipay: The real reason I'm doing it is for the last line there where we actually call the function. Command handlers always have to assume they are being passed in a data arg, but since healthcheck style commands take no arguments, its kind of clunky to force users to include a dummy arg in their custom healthcheck functions a la `def my_healthcheck(_)`. With this branch they can just do `def my_healthcheck()`. Don't know if this is the best way to do it though, I'm open to suggestions

languages/python/atom/element.py
plipay: Great idea! Went ahead with the former (dedicated "version" command auto-implemented by each element), its working well. I prefer this to a separate feature-tracking element, as that would introduce a 2nd point of failure on the system as a whole (in addition to the nucleus itself).

plipay:
Updated specification docs, added support for version/language checking, and verified that `wait_for_elements_healthy` works correctly for legacy elements that don't implement the healthcheck interface yet. I think this is good to merge unless you guys have any other callouts.

languages/python/atom/config.py
```diff
@@ -1,11 +1,14 @@
-VERSION = "0.1.0"
+VERSION = "0.10"
```
dpipemazo: this should likely get bumped I imagine? All of the old SDK clients will be identifying as `0.1.0`

Issue #158, opened 2019-06-05
Better multithreading support in Element


plipay:
Currently we use 2 redis pipelines, 1 for reads, 1 for writes. This will work if you have a writer/reader thread, but you can still run into concurrency issues if you have multiple reader/writer threads. Should circle back around to this and consider a better solution, such as a pool of pipelines.

PR #168, opened 2019-06-26
Support for multi-threaded use of send_command

plipay:
Fixes broken multithreading support in send_command, which was making a bunch of assumptions about the outcome of its xread calls. 

Use pool of pipelines instead of a shared write pipeline, as redis pipelines are not thread-safe.
https://github.com/elementary-robotics/atom/issues/158


languages/python/atom/element.py
```diff
@@ -29,32 +30,38 @@ def __init__(self, name, host=None, port=DEFAULT_REDIS_PORT, socket_path=DEFAULT
...
                 self._rclient = redis.StrictRedis(unix_socket_path=socket_path)
-            self._pipe = self._rclient.pipeline()
-            self._write_pipe = self._rclient.pipeline()
-            self._pipe.xadd(
+
+            # Init our pool of redis clients/pipelines
+            for i in range(REDIS_PIPELINE_POOL_SIZE):
+                self._rpipeline_pool.put(self._rclient.pipeline())
+
+            _pipe = self._rpipeline_pool.get()
```
dpipemazo: seems `block=True`, `timeout=None` on the get are the defaults so this will work; what are your thoughts on expressly codifying this in case the behavior changes? Pretty unlikely I imagine and likely not necessary?

languages/python/atom/element.py
```diff
@@ -98,6 +105,17 @@ def __del__(self):
         except redis.exceptions.RedisError:
             raise Exception("Could not connect to nucleus!")
 
+    def _release_pipeline(self, pipeline):
```
dpipemazo: :+1: nice

languages/python/atom/element.py
plipay: In this case, I prefer leaving them off because we make this call repeatedly and it ends up being a lot of extra text. I think its unlikely for the queue semantics to change (our python should be version locked anyway). If they did, we would find out pretty fast

PR #172, opened 2019-06-27
Python support for retrieving available commands of elements in atom-cli

GerardMaggiolino:
Support for #62 

Implements checking of available commands through atom-cli with the same front facing interface of "list streams". In atom-cli, "list commands" displays all available commands for all elements, while "list commands <element>" displays available commands for a single element. 

Retrieval of an element's commands uses a custom implemented command, similar to health and version checks, to prevent creating additional Redis structures. 

Retrieval of all commands at once is achieved by iterating through all existing Elements, sending a command to each. As a side-effect, for any Elements that do not have an active command loop, a failed to receive acknowledge is logged, as that Element will not be responding to commands. For the purpose of atom-cli, this is bypassed by not attempting to retrieve commands of the calling Element; however, if it's likely that other elements will have an inactive command loop, this implementation should be restructured. 

Two test cases have been created for testing both the custom command and the get all commands functionality. Additionally, in test_atom.py, pytest fixtures for yielding responders now explicitly invoke the garbage collector. Failing to do so while calling del on the element only de-increments the reference count of the object, and does not guarantee deletion in-between test cases (which causes inconsistent and rare test failures). Please let me know if this change should be a separate commit!

### DIFF
--- a/utilities/atom-cli/atom-cli.py
+++ b/utilities/atom-cli/atom-cli.py
@@ -44,12 +44,13 @@ class AtomCLI:
                   help [<command>]"""),
 
             "cmd_list": cleandoc("""
-                Displays available elements or streams.
-                Can filter streams based on element.
+                Displays available elements, streams, or commands.
+                Can filter streams and commands based on element.
 
                 Usage:
                   list elements
-                  list streams [<element>]"""),
+                  list streams [<element>]
+                  list commands [<element>]"""),
 
             "cmd_records": cleandoc("""
                 Displays log records or command and response records.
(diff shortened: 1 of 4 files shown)


GerardMaggiolino:
Added a version and language check, supporting elements at and above 0.3.0 in Python. Also added a test for this in test_atom.py.

utilities/atom-cli/atom-cli.py
```diff
@@ -165,7 +166,9 @@ def cmd_list(self, *args):
         mode_map = {
             "elements": self.element.get_all_elements,
             "streams": self.element.get_all_streams,
+            "commands": self.element.get_all_commands
         }
+        print(args)
```
plipay: Is this print intentional?

plipay:
See my question about the print statement, otherwise this looks great!

utilities/atom-cli/atom-cli.py
GerardMaggiolino: That's my bad, I forgot to remove this statement. This is now fixed.

PR #181, opened 2019-08-07
Added Version and Healthcheck support to C++ SDK

plipay:
- Adds version checking to C++ SDK, same as in the python SDK
- Adds health-checking to C++ SDK, same as in the python SDK
- Fixes bug in C SDK which would cause sendCommand attempts to non-existent elements to hang instead of returning NO_ACK errors
- Updates python SDK to allow healthchecking against C++ elements

### DIFF
--- a/languages/cpp/src/element.cc
+++ b/languages/cpp/src/element.cc
@@ -386,15 +386,23 @@ void Element::getElementVersion(
 	std::string element_name)
 {
 	std::map<std::string, msgpack::object> res;
-	enum atom_error_t err = this->sendCommandNoReq<std::map<std::string, msgpack::object>>(
+	enum atom_error_t err = this->sendCommand(
 		response,
 		element_name,
 		ATOM_VERSION_COMMAND,
-		res
+		NULL,
+		0
 	);
 	if (err != ATOM_NO_ERROR) {
 		return;
 	}
+	// Deserialize version command response
+	msgpack::object_handle oh = msgpack::unpack(
+		response.getData().c_str(),
+		response.getDataLen());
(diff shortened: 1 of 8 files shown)


languages/cpp/src/element.cc
```diff
@@ -531,6 +694,13 @@ void Element::addCommand(
 {
 	std::cout << "Creating command with name " << name << std::endl;
 
+	if ((name == ATOM_VERSION_COMMAND && commands.find(name) != commands.end()) ||
+		(name == ATOM_HEALTHCHECK_COMMAND && commands.find(name) != commands.end()))
+	{
+		error("Attempting to add reserved command name, choose a different command name");
```
dpipemazo: :+1:

dpipemazo: maybe subfunction this out as `checkCommandNameValid` or something since it's duplicated twice. Usually not a good idea to have this sort of thing in two cases for if/when we change it or add a list of protected names. I think the longest-term solution would be to subfunction it and make it be based off of a compile-time list of protected names s.t. when we go to add a new protected name it's just a one-line modification.

languages/cpp/test/test_element.cc
```diff
@@ -470,9 +470,32 @@ class MsgpackNoReqNoRes : public CommandMsgpack<std::nullptr_t, std::nullptr_t>
 	}
 };
 
+bool test_healthcheck_callback(
```
dpipemazo: :+1: thanks for writing tests!

dpipemazo:
Closing this out as it's out-of-date and we'll come back to it.

PR #193, opened 2019-09-16
190/c++ expose remote

dpipemazo:
Allows us to connect to remote redis using C++ API


dpipemazo:
Thanks @Cody-G :smile: . Going to close this out as it's quite old and we'll come back to it.

PR #194, opened 2019-09-19
Modify and test the default behavior of entry_read_since (reads new e…

### LINKED ISSUE #192: Fix error of usage of entry_read_since in docs
The docs state that
```If last_id is not passed, this call will return the first new piece of data that's been written after our call.```
However, this is not true as if `last_id` is not passed, we return the first piece of data in redis.

If we want to get the latest piece of data since the call, we should set `last_id = element._get_redis_timestamp()`

### PR DESCRIPTION
**Pull request by Cody-G**
…ntries only).  Fixes #192.

### DIFF
--- a/languages/python/tests/test_atom.py
+++ b/languages/python/tests/test_atom.py
@@ -2,7 +2,7 @@ import pytest
 import time
 import gc
 from atom import Element
-from multiprocessing import Process
+from multiprocessing import Process, Queue
 from threading import Thread, Lock
 from atom.config import ATOM_NO_ERROR, ATOM_COMMAND_NO_ACK, ATOM_COMMAND_UNSUPPORTED
 from atom.config import ATOM_COMMAND_NO_RESPONSE, ATOM_CALLBACK_FAILED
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: languages/python/tests/test_atom.py [hunk 1]
```diff
@@ -210,8 +210,12 @@ def test_read_since(self, caller, responder):
         for i in range(5):
             responder.entry_write("test_stream", {"data": i})
 
-        # Ensure this gets all entries
+        # Ensure this doesn't get an entry (because it's waiting for new entries and they never come)
         entries = caller.entry_read_since("test_responder", "test_stream")
+        assert(len(entries) == 0)
+
+        # Ensure this gets all entries
+        entries = caller.entry_read_since("test_responder", "test_stream",last_id='0')
```
cheripai: Can we add a test to validate that `last_id="$"` works? 
Thanks :grin:

Cody-G: Sure! Let me know if the test I just added does the trick.

cheripai: Looks good, thanks!

Issue #60, opened 2018-12-11
Use official redis-py client


cheripai:
The official python redis client supports streams now

cheripai:
:open_mouth:

PR #196, opened 2019-09-20
Cjg/new redis py

Cody-G:
This should address #60.  Most of the breaking changes are explained in the redis-py readme [here](https://github.com/andymccurdy/redis-py).  I didn't make an effort to shield atom users from the breaking changes.  You can get an idea of how usage will change by checking out the changes to the test code.  If people prefer the old behavior then we can add more code to process redis-py inputs and outputs internally. However I think that complicates the code base, and may keep getting worse if redis-py continues to change.

Note we'll need to rebase before merging, I'll get #194 merged first.


Cody-G:
This is good to go again.  I'll continue to rebase as needed with other PRs landing.

Cody-G:
> Let's change these APIs where we can't pass None on the user side 

Done!

PR #203, opened 2019-10-06
python: add side-channel for raw data

dpipemazo:
This will allow us to msgpack main commands but still
send additional data (such as an image) over the raw data
side-channel and it will be passed as a kwarg to the
handler.

Addition of new side-channel causes no code changes to
existing elements


Cody-G:
What do you think about also modifying the Response class to support unserialized side channel data? This would be more symmetric, and there's an immediate use case for image alignment: it's useful for the caller to receive an aligned image and also a score indicating the quality of the alignment. And/or a small matrix specifying what transform was found.

dpipemazo:
@Cody-G I think for now we can hold off but it's an idea worth considering

PR #207, opened 2019-10-23
python: remove try/except from user handler

dpipemazo:


dpipemazo:
@cheripai I think letting it time out and then checking the logs is OK for now, though it's a valid point. Yes, let's merge it in

PR #212, opened 2019-11-06
Python: Implement reference APIs

dpipemazo:
Five new APIs:
1. reference_create -- makes a new reference. Supports msgpack
2. reference_get -- Gets data from a reference. Supports msgpack
3. reference_delete -- deletes a reference
4. reference_update_timeout -- updates the timeout for a reference
5. reference_get_timeout_ms -- Returns how long until timeout on a
	reference

Docs to follow


languages/python/atom/element.py
```diff
@@ -708,3 +719,150 @@ def log(self, level, msg, stdout=True):
...
+
+        # Now, we can go ahead and do the SET in redis for the key
+        _pipe = self._rpipeline_pool.get()
+        px_val = timeout_ms if timeout_ms != 0 else None
+
+        # Do the set. Expire as set by the user, throw an error if the
+        #   value already exists (since this would be a collision on the UUID)
+        _pipe.set(key, data, px=px_val, nx=True)
+        _pipe.execute()
+        _pipe = self._release_pipeline(_pipe)
```
dpipemazo: error-check this

languages/python/atom/element.py
```diff
@@ -708,3 +719,153 @@ def log(self, level, msg, stdout=True):
...
+
+        # Make sure there's only one value in the data return
+        if type(data) != list and len(data) != 1:
+            raise ValueError(f"Invalid response from redis: {data}")
+
+        if data[0] != 1:
+            raise KeyError(f"Reference {key} not in redis")
+
+    def reference_update_timeout(self, key, timeout_ms):
```
Cody-G: Minor nitpick:  Could we match the name of this function better with the update function?  We could either name it `reference_update_timeout_ms` or remove the `_ms` suffix from `reference_get_timeout_ms`.

dpipemazo: Yes, this is a good ask. Will clean it up.

dpipemazo:
A few decently big improvements have been worked in:

1. New reference_create_from_stream API. This allows us to specify an element, stream and (optional) stream ID and create a set of references from it. An independent reference will be created for each key in the stream, where the reference will end with `:key`. This is nice if we have cameras publishing large k:v pairs on a single stream -- we can access/treat them individually.

2. In order to accomplish (1) we have our first lua script built into atom! This is pretty cool -- you can load a lua script into redis and then call it with a set of arguments. This is nice so that for more complex commands such as `reference_create_from_stream_id` we can guarantee the same performance across language clients. It also most importantly allows us to never have the large image data for image references leave redis -- a caller can make a reference for an image without the image needing to be copied to the caller and then back to redis in the `create_reference` command.

3. To work with (2), we now have a top-level `lua-scripts` folder. This is where we can store shared lua scripts that will be used across all of the language clients. It is expected that each client loads the script on element init. We use the SCRIPT LOAD command to load a lua script and get a SHA back from redis that identifies the script. Then, we can call EVALSHA on the return of the SCRIPT LOAD to call the script without having to transmit it back/forth each time. This is best for performance but does introduce an interesting quirk into the system: Each client loads its own version of the lua scripts when it boots. This is nice -- it allows us to change/modify the lua scripts and as long as we change the libraries to go along with them older clients with older scripts can still work. It's not necessarily how I wanted to do it -- the other way would be to do it globally for everyone in the nucleus -- but redis doesn't support loading a script under some global name like a function.

4. To accomplish (3) we now need to ship a lua script with the python client and make it work not only in docker containers but on any install of our python library. Added a `MANIFEST.in` and installed the script along with the package so it should work on all installs and should also be OS-portable.

Docs still to follow before merging, added some decent unit test coverage.

PR #221, opened 2019-11-08
reference_get_list: initial addition

dpipemazo:
Add API to get list of references and return a dictionary of
references mapped to their data values


Cody-G:
For symmetry should we consider adding reference_create_list?

dpipemazo:
@Cody-G yeah I was thinking about this -- I think it's pretty rare that at the same time you're going to want to be creating multiple different references. I'd rather let it sit and if there is a valid use case happy to add it.

dpipemazo:
@Cody-G in talking to Dat it is indeed a use case, will add it :+1:

PR #223, opened 2019-11-09
220/reference get list of keys

cheripai:
This provides the ability to work with one to many references without
creating additional functions.

reference_create returns a list of reference_ids
reference_get returns a dictionary of reference_id : data


dpipemazo:
Nice! :tada:

PR #229, opened 2019-11-18
Change return type of reference_get()

Cody-G:
Returns a list now rather than a dictionary


dpipemazo:
Looks good, merge when you're ready

PR #234, opened 2020-01-21
Issue #230 - Integrate Apache Arrow serialization

### LINKED ISSUE #230: RFC: Integrate Apache Arrow as a serialization option
@Cody-G has pointed out a few use cases where our current serialization options aren't optimal -- specifically large, structured arrays/matrices that aren't necessarily images. In these instances `msgpack` is [not great](http://satoru.rocks/2018/08/fastest-way-to-serialize-array/) and [Apache Arrow](https://arrow.apache.org/) is much better. Arrow's language support, though not at the same level as Redis's or MsgPack's is quite solid and would likely cover the majority of use cases in which one would be processing such arrays.

Interested to hear thoughts on:
1. Adding Arrow as a serialization option
2. If (1), then should we change the serialization args from T/F to one of a list of options, such as in Python, changing from:
```
entry_write(..., serialize=True)
```
to
```
entry_write(..., serialization="msgpack")
```
or similar

aywaldron:
- Adds str `serialization` option to Element and Response API; defaults to `msgpack`
- Changes boolean `msgpack` CLI setting to str `serialize`; defaults to `msgpack`
- Adds Apache Arrow as serialization option; installs `pyarrow` in atom and nucleus docker images
- Pulls serialization functionality into new `atom.serialization` module
- Adds `"ser"` key for serialization method to entries; uses its value to deserialize entries on read
- Adds serialization information to reference keys and uses for deserialization of references
- Adds serialization method to command handler map
- Preserves legacy functionality of using msgpack serialization when `(de)serialize=True` and no other serialization setting found
- Removes side channels for raw data from command and response in favor of using references
- Updates documentation and tests with all changes

Resolves #230


languages/python/atom/serialization.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class arrow():
+    """
+    Class containing Apache Arrow serialization and deserialization functions.
+    """
+
+    @classmethod
+    def serialize(cls, data):
+        return pa.serialize(data).to_buffer().to_pybytes()
```
Cody-G: `to_pybytes` makes an additional copy (see docs [here](https://arrow.apache.org/docs/python/memory.html)).  I'm pretty sure we can avoid calling that method.  My guess at one way to do it:

```python
@classmethod
def serialize(cls, data):
    data_serialized = pa.serialize(data)
    return memoryview(data_serialized.to_buffer())
```

The reason I'm guessing this will work is that the redis-py package communicates with Redis over a socket (see [this](https://github.com/andymccurdy/redis-py/blob/ff69f0d77284643909462ee6d1e37233c6677672/redis/connection.py#L686) line) and sockets support memoryviews (and anything else that implements Python's buffer API).  Happy to talk more in person -- my code is a guess and may not work without some tweaks.

I'm also wondering whether we're better off using `pa.serialize(data).to_components()` and `deserialize_components(data_serialized)`.  I know at the design review meeting we said don't worry about it, but on second thought it seems like we'd be getting better performance for nested objects almost _for free_.  What do you think?

doc/source/includes/walkthrough.md
```diff
@@ -418,10 +418,11 @@ Now that our atombot element is running, let's interact with it using `atom-cli`
 - We can control the rate at which the messages are printed by passing in the rate value (hz)
     - `read atombot pos_map 1`
 - By default, atom-cli automatically msgpacks the data that is sent/received
-- If you wish, you can override this with the `msgpack` command
-    - `msgpack False`
+- If you wish, you can override this with the `serialize` command
+    - `serialize none`
```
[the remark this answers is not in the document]
dpipemazo: +1 on this -- though the `serialize` boolean should now be more or less deprecated

languages/python/atom/element.py
```diff
@@ -334,21 +334,22 @@ def get_all_commands(self, element_name=None, ignore_caller=True):
                 command_list.extend([f'{element}:{command}' for command in elem_commands])
         return command_list
 
-    def command_add(self, name, handler, timeout=RESPONSE_TIMEOUT, deserialize=False):
+    def command_add(self, name, handler, timeout=RESPONSE_TIMEOUT, deserialize=False, serialization="msgpack"):
```
dpipemazo: I think we should make the `deserialize` kwarg be the final kwarg in the function call since it's now legacy/deprecated.

dpipemazo: I now see how this is being used -- I think a good default should be `serialization=None` across the board. The default is then no serialization and if the value is non-None we can go ahead and interpret the string. This would help us get rid of the two-argument structure we have here and fully deprecate the `deserialize` flag.

languages/python/atom/element.py
```diff
@@ -334,21 +334,22 @@ def get_all_commands(self, element_name=None, ignore_caller=True):
...
-            deserialize (bool, optional): Whether or not to deserialize the data using msgpack before passing it to the handler.
+            deserialize (bool, optional): Whether or not to deserialize the data before passing it to the handler.
+            serialization (str, optional): If deserializing, the method of serialization to use; defaults to msgpack.
         """
         if not callable(handler):
             raise TypeError("Passed in handler is not a function!")
         if name in self.reserved_commands and name in self.handler_map:
             raise ValueError(f"'{name}' is a reserved command name dedicated to {name} commands, choose another name")
-        self.handler_map[name] = {"handler": handler, "deserialize": deserialize}
+        self.handler_map[name] = {"handler": handler, "deserialize": deserialize, "serialization": serialization}
```
dpipemazo: I don't think we're currently error-checking this. We should identify and raise if an invalid serialization method is passed.

languages/python/atom/element.py
```diff
@@ -455,9 +448,10 @@ def command_loop(self):
                     err_code=ATOM_COMMAND_UNSUPPORTED, err_str="Unsupported command.")
             else:
                 if cmd_name not in self.reserved_commands:
-                    data = unpackb(
-                        data, raw=False) if self.handler_map[cmd_name]["deserialize"] else data
-                    response = self.handler_map[cmd_name]["handler"](data, **kw_data)
+                    if self.handler_map[cmd_name]["deserialize"]:
+                        # assume serialization key present if deserialize set to True
+                        data = ser.deserialize(data, method=self.handler_map[cmd_name]["serialization"])
```
dpipemazo: I think this will throw an error if the serialization method isn't error-checked in the `command_add` function

languages/python/atom/element.py
```diff
@@ -498,8 +494,9 @@ def command_send(self,
             cmd_name (str): Name of the command to execute of element_name.
             data: Entry to be passed to the function specified by cmd_name.
             block (bool): Wait for the response before returning from the function.
-            serialize (bool, optional): Whether or not to serialize the data using msgpack before sending it to the command.
-            deserialize (bool, optional): Whether or not to deserialize the data in the response using msgpack.
+            serialize (bool, optional): Whether or not to serialize the data before sending it to the command.
+            deserialize (bool, optional): Whether or not to deserialize the data in the response.
+            serialization (str, optional): If (de)serializing, the method to use; defaults to msgpack
```
dpipemazo: This is interesting, since we have serialization on the command sending side as well as the command receiving side. Assuming no legacy behavior, we should only need to specify the serialization on the send-side. I think the comment should be updated to reflect this. The booleans should be left for legacy purposes and documented as such

languages/python/atom/element.py
```diff
@@ -576,33 +574,38 @@ def command_send(self,
...
-
-                    # Process the raw data
-                    raw_data = {}
-                    for val in response:
-                        ascii_val = val.decode('ascii')
-                        if ascii_val not in RES_RESERVED_KEYS:
-                            raw_data[ascii_val] = response[val]
+                    # check response for serialization method; if not present, use user specified method
+                    if b"ser" in response:
+                        ser_method = response.get(b"ser").decode()
```
dpipemazo: Is the `.get` function necessary/correct if you've already checked to see if it's in the response? I've seen it used with `response.get(b"ser", None)` to do both the get and check in a single line, and then you check the return of that.

languages/python/atom/element.py
```diff
@@ -649,11 +653,18 @@ def entry_read_loop(self, stream_handlers, n_loops=None, timeout=MAX_BLOCK, dese
                 for uid, entry in msgs:
                     streams[stream] = uid
                     entry = self._decode_entry(entry)
-                    entry = self._deserialize_entry(entry) if deserialize else entry
+                    if "ser" in entry.keys():
+                        ser_method = entry.pop("ser").decode()
```
dpipemazo: Is the pop necessary? It might be since I think we might be then converting the resulting dictionary into an object -- just wanted to confirm.

languages/python/atom/element.py
```diff
@@ -889,8 +926,18 @@ def reference_get(self, *keys, deserialize=False):
         if type(data) is not list:
             raise ValueError(f"Invalid response from redis: {data}")
 
-        if deserialize:
-            return [unpackb(v, raw=False) if v is not None else None for v in data]
+        # look for serialization method in reference key first; if not present use user specified method
+        key_split = key.split(':') if type(key) == str else key.decode().split(':')
```
dpipemazo: I don't think this currently handles different serialization methods for different keys? We should add a test case for this. If I'm reading this correctly I think we only are using the deserialization method from the final key and applying it to all keys. Each key should be able to independently support its own serialization method

languages/python/atom/messages.py
```diff
@@ -33,35 +37,35 @@ def __init__(self, element, cmd, data, **kwargs):
...
             raise TypeError("err_str must be a str")
-        if any(key in raw_data for key in RES_RESERVED_KEYS):
-            raise KeyError("invalid key in raw_data")
-        self.data = packb(data, use_bin_type=True) if serialize else data
-        self.__dict__.update(raw_data)
+
+        self.data = ser.serialize(data, method=serialization) if serialize else data
         self.err_code = err_code
         self.err_str = err_str
+        if serialize:
```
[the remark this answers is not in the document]
dpipemazo: I like using `none` to indicate no serialization, else having a string to indicate the type of serializaton

languages/python/atom/serialization.py
```diff
@@ -0,0 +1,100 @@
...
+                            "to pickle. Change data type or choose a different serialization method.")
+        else:
+            return pa.serialize(data).to_buffer().to_pybytes()
+
+    @classmethod
+    def deserialize(cls, data):
+        return pa.deserialize(data)
+
+
+class Serializations(Enum):
```
dpipemazo: Perhaps add an `is_valid` function to this class that can be used to validate serialization strings from the user as mentioned in some of the previous comments.

languages/python/atom/serialization.py
```diff
@@ -0,0 +1,100 @@
...
+    Args:
+        data: The data to serialize.
+        method (str, optional): The serialization method to use; defaults to msgpack
+    Returns:
+        The serialized data.
+    Raises:
+        ValueError if requested method is not in available serialization options defined
+        by Serializations enum.
+    """
+    if method not in Serializations.__members__:
```
dpipemazo: ah, this would be the the `is_valid` function on the `Serializations` class. Would be good to expose this/put it in one place since I think I see it a few times in this file.

languages/python/atom/serialization.py
```diff
@@ -0,0 +1,100 @@
+from msgpack import packb, unpackb
+import pyarrow as pa
+from enum import Enum
+import builtins
+
+
+class msgpack():
```
dpipemazo: I believe standard/our convention in Python is to use CamelCase on class names. We follow the [Google Style Guide](http://google.github.io/styleguide/pyguide.html) for the most part.

languages/python/atom/serialization.py
```diff
@@ -0,0 +1,100 @@
+from msgpack import packb, unpackb
+import pyarrow as pa
+from enum import Enum
+import builtins
+
```
dpipemazo: I quite like the class-based definitions in this file. One thought might be that we want to define a generic SerializationMethod class which all of the other classes inherit from/override the functionality. The generic SerializationMethod class could implement the "none" type serialization, i.e. just be a pass-through. It would help clean up the branching logic of "if serialization, else" throughout the codebase by just always calling the class.

lua-scripts/stream_reference.lua
```diff
@@ -15,18 +15,31 @@ else
     data = redis.call('xrange',ARGV[1],ARGV[2],ARGV[2])
 end
 
--- Loop over the response. We will ket a sequence of key, value pairs
+-- Loop over the response. We will get a sequence of key, value pairs
```
dpipemazo: :slightly_smiling_face:

lua-scripts/stream_reference.lua
```diff
@@ -15,18 +15,31 @@ else
...
 --  in order. We tell the difference between them by checking the odd/even
 --  of the iterator. If the iterator is odd then it's a key, else a value.
 -- For each key,value pair, make a reference at reference:id:key with the
 --  value as the value.
 local ref = ""
+local ser = ""
 local keys = {}
+
+-- Find serialization method from "ser" key if it exists
+for key,val in pairs(data[1][2]) do
```
dpipemazo: do we need to do this in a separate loop? I imagine the performance of looping twice is _slightly_ worse than just doing this in the same loop as we get to the key?

dpipemazo:
Also, please add your name to the list of contributors in the docs before this gets merged in :slightly_smiling_face:

Cody-G:
>I think I'm personally in favor of dropping the serialize and deserialize kwargs from the API altogether and just focusing on the new serialization key.

+1 for Dan's suggestion of dropping these altogether, tagging the current master (before this PR) to smooth out the transition for existing Elements.  This simplifies the code for both Atom and Elements a lot, and I think we're agile enough that the transition won't be too bad.

> The current default is no serialization which I think I might prefer. It would make our code more wordy in many places, but it would also make explicit what's going on under the hood which I think would add value. 

+1 for making the default None.  The transparency is nice, and it doesn't commit us to one serialization method more than another.  (Also Element code will still be less verbose than it was before this PR)

> Every command_add call would be forced to explicitly specify the serialization

Since we're already doing this it seems that we could already remove the requirement that the caller specifies the serialization method.  I know this came up as a nice-to-have in the design review, after further though I'm wondering if it's worthwhile to just do this now?  It removes a lot of code complexity, and simplifies the documentation of Atom and all of our Elements -- after that change Element commands would work like normal functions for the most part.

Dockerfile-nucleus
```diff
@@ -22,6 +22,54 @@ RUN git clone https://github.com/msgpack/msgpack-c.git
...
+RUN pip3 install --no-cache-dir six pytest numpy cython
+ARG ARROW_VERSION=0.15.1
+ARG ARROW_BUILD_TYPE=release
+ENV ARROW_HOME=/usr/local \
+    PARQUET_HOME=/usr/local
+ENV PYTHON_EXECUTABLE=/usr/bin/python3
+
+RUN mkdir /arrow \
+    && apk add --no-cache curl \
+    && curl -o /tmp/apache-arrow.zip -SL https://codeload.github.com/apache/arrow/zip/master \
```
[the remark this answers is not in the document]
aywaldron: Sounds good -- I'll add a submodule in `third-party` with an install script. I used master for the time being since their latest releases don't build on alpine but using the most recent version that does should be sufficient.

doc/source/includes/sdk-concepts.md
```diff
@@ -17,7 +17,7 @@ Issued by an element to execute some functionality of another element.
 Returned by an element to indicate the results of the command to the caller element.
 
 ## Entry
-A timestamped data packet that is published by an element on a stream that can contain multiple fields of data. The atom system is not concerned with the serialization of the data and leaves it as the responsibility of the developers to know how the data was serialized in the element. Our recommendation for serialization is msgpack as it is supported by the major programming languages.
+A timestamped data packet that is published by an element on a stream that can contain multiple fields of data. The atom system will create a "ser" key in the entry with the serialization method, set to "none" if no serialization was used. It will then use this key to deserialize the entry when reading it. The default serialization method used by Atom is msgpack as it is supported by the major programming languages. Apache Arrow should be used for array-like data.
```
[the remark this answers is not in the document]
aywaldron: Sure.

lua-scripts/stream_reference.lua
aywaldron: It's in a separate loop because we do not know where in the list it will be (it could be the last pair), and we need its value for setting the keys of the other entries. We have to find its value first and then loop over the remaining pairs, so I'm not sure how to avoid looping twice.

doc/source/includes/sdk-specification.md
```diff
@@ -403,7 +405,9 @@ enum atom_error_t err = element.entryReadSince(
 # This will get the 10 oldest entries from your_stream since the beginning of time.
 entries = my_element.entry_read_since("your_element", "your_stream", last_id="0", n=10)
 
-# If the element is publishing serialized entries, they can be deserialized
+# If the element is publishing serialized entries, they can be deserialized.
+# The entry will be checked for a "ser" flag to determine the deserialization method.
+# If not present, the serialization option will be used for deserialization, defaulting to msgpack.
 entries = my_element.entry_read_since("your_element", "your_stream", last_id="0", n=10, deserialize=True)
```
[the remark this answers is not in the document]
aywaldron: I may have misunderstood the desired functionality with the legacy `serialize`/`deserialize` and new `serialization` parameters. I've written everything to require `serialize`/`deserialize` to be True in order to use `serialization`. All of your comments are correct for fully deprecating `serialize`/`deserialize`.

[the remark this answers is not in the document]
plipay: Sort of. I think serialize=True needs to be the case for serialization to happen on the stream/command on the sender side. However, on deserialization, I think the default value for deserialize should be True, such that setting deserialize=True is a no-op. Setting deserialize=False would explicitly skip doing any deserialization and be equivalent to if the message had been sent with a serialization type of None.

doc/source/includes/sdk-specification.md
```diff
@@ -751,22 +757,15 @@ def add_1(data):
 my_element.command_add("add_1", add_1, timeout=50)
 
 # Alternatively, we could use serialization to keep from having to convert data types.
+# The serialization method to use can be specified using the serialization option. If
+# serialize is set to True and serialization is not specified, msgpack will be used.
 def add_2(data):
-    return Response(data + 2, serialize=True)
+    return Response(data + 2, serialize=True, serialization="msgpack")
```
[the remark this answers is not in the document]
plipay: You still need the serialize key for legacy reasons right? Or are you saying keep the key but get rid of it in the docs?
I thought it was:
                                                                    _key on msg stream_
 serialize=True, serialization="msgpack" ->     "msgpack"
 serialize=True, serialization="arrow"       ->     "arrow"
 serialize=False                                        ->     "none"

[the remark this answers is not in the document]
aywaldron: ^ That's how its currently written. I think Dan is suggesting we do not require `serialize=True` in order to use `serialization`.

aywaldron:
As written, the default serialization is None, since its looking for the `serialize` flag to be True before checking the `serialization` method. It sounds like this is not the functionality we want. Fully deprecating the `serialize` flag would simplify the current code somewhat. Removing it altogether would make it much simpler.

languages/python/atom/element.py
aywaldron: Yes it will, good catch

languages/python/atom/serialization.py
```diff
@@ -0,0 +1,100 @@
...
+        The serialized data.
+    Raises:
+        ValueError if requested method is not in available serialization options defined
+        by Serializations enum.
+    """
+    if method not in Serializations.__members__:
+        raise ValueError(f'Invalid serialization method. Must be one of {Serializations.print_values()}.')
+
+    if Serializations[method].value:
+        return Serializations[method].value.serialize(data)
```
plipay: This is very nifty, I didn't know you could use classes as value types on Enums. When possible with None checks, try to explicitly do `if Serializations[method].value is not None`. I'm guilty of this shortcut sometimes too, but trying to not do it anymore as there are cases where python will get you in trouble with the generic check

languages/python/atom/element.py
aywaldron: I thought we needed this because an element might receive legacy type commands where it's necessary to specify deserialization. If not then yes this should be updated.

languages/python/atom/element.py
aywaldron: You're right, it's not necessary

utilities/atom-cli/atom-cli.py
```diff
@@ -80,12 +80,13 @@ def __init__(self):
                 Usage:
                   exit"""),
 
-            "cmd_msgpack": cleandoc("""
-                Turns on/off msgpack serialization and deserialization.
-                Pass True to use msgpack, False to turn off. Defaults to True.
+            "cmd_serialize": cleandoc("""
+                Sets serialization/deserialization setting to either use msgpack,
+                Apache arrow, or no (de)serialization. Defaults to no serialization.
```
plipay: I think the current behavior is msgpack serialization on send is enabled by default. I think you might want to make that your default as well, since most of our services still use msgpack serialization, so the first thing devs are going to be doing when they open cli is generally re-enabling it.

languages/python/atom/element.py
aywaldron: Yes, the rest of the entry is first deserialized and then passed to the stream handler. I assumed we do not want the "ser" key to be included with that.

languages/python/atom/element.py
aywaldron: Yes you're correct. Will update.

lua-scripts/stream_reference.lua
dpipemazo: ah, this makes sense :+1:

languages/python/atom/serialization.py
aywaldron: I used `to_pybytes` because Redis throws the following error on any data type that is not a bytes, str, int, or float (including Arrow buffers and memoryviews):
```
redis.exceptions.DataError: Invalid input of type: 'memoryview'. Convert to a byte, string or number first.
```
and I didn't see any way around this. 

Trying to send the full dictionary from `to_components` gives the same error. From my understanding, you would have to convert every `pyarrow.Buffer` object returned by `to_components` to bytes individually for Redis to accept them, and I'm not sure if that gives us better performance or not.

third-party/install_arrow.sh
```diff
@@ -0,0 +1,58 @@
+# Script to build and install arrow on alpine
+
+# build Cody's fork of redis-py
+git clone https://github.com/Cody-G/redis-py.git
```
dpipemazo: We should fork/host this on elementary robotics and this should be a submodule like arrow. The goal with this is to be in as much control as possible over our build process so that we can minimize breakage of the build

dpipemazo: `redis-py` now forked and hosted on ER: https://github.com/elementary-robotics/redis-py

aywaldron: @Cody-G opened a [PR on redis-py](https://github.com/andymccurdy/redis-py/pull/1266) that he is working on and seems like will be merged in. This was a temporary test and I was going to wait for redis-py to be updated with his change. Do you still prefer using our own fork?

Cody-G: There's some work to be done there and it's possible that the pythone 2.7 support won't be simple. How about we go with our fork for now and I can make a PR if that gets merged later?

doc/source/includes/sdk-concepts.md
```diff
@@ -17,7 +17,13 @@ Issued by an element to execute some functionality of another element.
 Returned by an element to indicate the results of the command to the caller element.
 
 ## Entry
-A timestamped data packet that is published by an element on a stream that can contain multiple fields of data. The atom system is not concerned with the serialization of the data and leaves it as the responsibility of the developers to know how the data was serialized in the element. Our recommendation for serialization is msgpack as it is supported by the major programming languages.
+A timestamped data packet that is published by an element on a stream that can contain multiple fields of data. The atom system will create a "ser" key in the entry with the serialization method, set to "none" if no serialization was used. It will then use this key to deserialize the entry when reading it. By default, Atom will not serialize any data. Serialization must be specified using the "serialization" option in the API. Below are the available serialization methods.
```
dpipemazo: Thanks for adding this section on serialization. I think it branches a bit deeper than just on the entry and warrants its own section, with notes calling out to it in the following sections:
- Entry
- Command/Response
- Reference

languages/python/atom/element.py
```diff
@@ -329,26 +330,40 @@ def get_all_commands(self, element_name=None, ignore_caller=True):
...
             if self._check_element_version(element, {'Python'}, 0.3):
                 # Retrieve commands for each element
-                elem_commands = self.command_send(element, COMMAND_LIST_COMMAND, deserialize=True)['data']
+                elem_commands = self.command_send(element, COMMAND_LIST_COMMAND, serialization="msgpack")['data']
                 # Rename each command pre-pending the element name
                 command_list.extend([f'{element}:{command}' for command in elem_commands])
         return command_list
 
-    def command_add(self, name, handler, timeout=RESPONSE_TIMEOUT, deserialize=False):
+    def command_add(self, name, handler, timeout=RESPONSE_TIMEOUT, serialization="none", deserialize=None):
```
dpipemazo: I think we should use a `None` type instead of the "none" string here. `None` would then be no serialization, then if a string, we check it against the serialization types.

languages/python/atom/serialization.py
```diff
@@ -0,0 +1,116 @@
+from msgpack import packb, unpackb
+import pyarrow as pa
+from enum import Enum
+import builtins
+
+
+class GenericSerializationMethod():
```
dpipemazo: :+1:

languages/python/atom/element.py
aywaldron: Ok sure. It will add a small amount of complexity in translating it to the string "none" for use in reference keys and checking enum membership. Currently the "none" string is passed all the way through the serialization module without requiring any checks.

aywaldron: It also brings up the question of if the string in redis should always be lowercase "none" or if capitalized "None" is also acceptable.

.gitmodules
```diff
@@ -4,3 +4,9 @@
 [submodule "third-party/docker-opengl"]
 	path = third-party/docker-opengl
 	url = https://github.com/thewtex/docker-opengl
+[submodule "third-party/apache-arrow"]
+	path = third-party/apache-arrow
+	url = https://github.com/apache/arrow
+[submodule "languages/python/third-party/redis-py"]
+	path = languages/python/third-party/redis-py
+	url = git@github.com:elementary-robotics/redis-py.git
```
dpipemazo: this should probably be an https clone for users without ssh keys.

languages/python/atom/element.py
```diff
@@ -498,9 +511,16 @@ def command_send(self,
...
             block (bool): Wait for the response before returning from the function.
-            serialize (bool, optional): Whether or not to serialize the data using msgpack before sending it to the command.
-            deserialize (bool, optional): Whether or not to deserialize the data in the response using msgpack.
-            ack_timeout (int, optional): Time in milliseconds to wait for ack before timing out, overrides default value
+            ack_timeout (int, optional): Time in milliseconds to wait for ack before
+                                         timing out, overrides default value.
+            serialization (str, optional): Method of serialization to use;
+                                           defaults to None.
+
+            Deprecated:
```
dpipemazo: :+1:

languages/python/atom/element.py
```diff
@@ -721,19 +780,31 @@ def entry_write(self, stream_name, field_data_map, maxlen=STREAM_LEN, serialize=
...
-            entry = Entry(serialized_field_data_map)
-        else:
-            entry = Entry(field_data_map)
+
+        if serialize is not None:  # check for deprecated legacy mode
+            serialization = "msgpack" if serialize else None
+
+        ser_field_data_map = {}
+        for k, v in field_data_map.items():
+            if k == "ser":
```
dpipemazo: should we make this more generic for other reserved keys in the future?

utilities/atom-cli/atom-cli.py
```diff
@@ -336,22 +344,24 @@ def cmd_read(self, *args):
...
+            print(f"\nPass one argument: {ser.Serializations.print_values()}.")
             return
 
-        # Otherwise try to get the new msgpack valud
-        if (args[0].lower() == "true"):
-            self.use_msgpack = True
-        elif (args[0].lower() == "false"):
-            self.use_msgpack = False
+        # Otherwise try to get the new setting
+        if (args[0].lower() == "msgpack"):
```
dpipemazo: this could potentially be done with a lookup in a dict instead of the if/elif/else. Would perhaps save a string compare or two.

Issue #237, opened 2020-01-24
Trigger builds from GitHub tags; push version tagged images


aywaldron:
Creating tags in GitHub should automatically trigger a build on Circle CI. As part of this build, a docker image tagged with the GitHub tag should be pushed to the registry along with the build number and latest tags.

aywaldron:
Closed by PRs #238 and #246

PR #238, opened 2020-01-24
Trigger CircleCI builds on GitHub tags (#237)

aywaldron:
When a GitHub tag is created, a master build will be triggered. The docker image will be pushed with the following tags:
- `master-<circle build number>`
- `latest`
- `<GitHub tag>`

### DIFF
--- a/.circleci/atom.yml
+++ b/.circleci/atom.yml
@@ -38,12 +38,16 @@ commands:
   # Load image from workspace and log into docker
   load_and_login:
     description: "Load the image from workspace and login"
+    parameters:
+      image_name:
+        type: string
+        default: image
     steps:
       - attach_workspace:
           at: /tmp/workspace
       - run:
           name: Load Docker Image
-          command: docker load -i /tmp/workspace/image.tar
+          command: docker load -i /tmp/workspace/<< parameters.image_name >>.tar
       - docker_login
 
   # Build and launch the docker-compose file
@@ -68,17 +72,20 @@ commands:
   # Store the docker image
   store_image:
     description: "Store the image built"
+    parameters:
+      image_filename:
+        type: string
+        default: image
     steps:
       - run:
           name: Archive Docker image
-          command: docker save -o image.tar ${CIRCLE_PROJECT_REPONAME}-${CIRCLE_WORKFLOW_ID}
+          command: docker save -o << parameters.image_filename >>.tar ${CIRCLE_PROJECT_REPONAME}-${CIRCLE_WORKFLOW_ID}
       - persist_to_workspace:
           root: .
           paths:
-            - ./image.tar
+            - ./ << parameters.image_filename >>.tar
       - store_artifacts:
-          path: ./image.tar
-
+          path: ./ << parameters.image_filename >>.tar
 
   # Create and push image tag
   tag_and_deploy:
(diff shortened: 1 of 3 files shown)


aywaldron:
New changes:
- Atom orb commands have been parameterized with defaults to increase flexibility.
- Atom orb jobs for pushing dev/master/latest tags have been moved to commands so they can be used more flexibly for atom's build of multiple containers. Another option is to keep them as jobs and then have separate deploy jobs for every image that Atom pushes. That would be 4 images x 3 types of deployment = 12 different jobs. 
- Rewrite of atom's build config to use new orb
  - Parameterize testing command 
  - Run tests in opengl, cuda, and opengl-cuda containers in addition to basic atom container
  - Parallelize build and test of opengl, cuda, and opengl-cuda containers
  - Use newly parameterized orb commands for docker login, build, launch, store, load, tag, and deploy

.circleci/atom.yml
```diff
@@ -54,45 +58,117 @@ commands:
...
     steps:
       - run:
           name: Archive Docker image
-          command: docker save -o image.tar ${CIRCLE_PROJECT_REPONAME}-${CIRCLE_WORKFLOW_ID}
+          command: docker save -o << parameters.image_filename >>.tar << parameters.image_tag >>
       - persist_to_workspace:
           root: .
           paths:
-            - ./image.tar
+            - .
```
aywaldron: Using `./<< parameters.image_filename >>.tar` here threw the error 
```
The specified paths did not match any files in /home/circleci/project
```
even though the .tar file was present in the directory. This works around that by persisting the project directory.

docker-compose-circle.yml
```diff
@@ -0,0 +1,43 @@
...
+        volume:
+          nocopy: true
+
+  atom:
+    container_name: dev_atom
+    build:
+      context: .
+      dockerfile: Dockerfile-atom
+      args:
+        BASE_IMAGE: $BASE_IMAGE
```
aywaldron: This docker compose file is used for building and testing the cuda, opengl, and opengl-cuda atom images since optional build args are required to specify a base image.

.circleci/config.yml
```diff
@@ -1,220 +1,347 @@
...
+  target_image: ${DOCKERHUB_ORG}/${DOCKERHUB_DOCS_REPO}
+
+
+#
+# Circle config
+#
+version: 2.1
+
+orbs:
+  atom: elementaryrobotics/atom@dev:237
```
aywaldron: The new orb is currently published with this tag.

.circleci/atom.yml
```diff
@@ -35,16 +35,31 @@ commands:
           name: Docker Login
           command: echo ${DOCKERHUB_PASSWORD} | docker login -u ${DOCKERHUB_USER} --password-stdin
 
-  # Load image from workspace and log into docker
+  # Attach workspace, load image, and log into docker
   load_and_login:
     description: "Load the image from workspace and login"
+    parameters:
+      image_filename:
```
dpipemazo: :+1: nice

.circleci/atom.yml
```diff
@@ -35,16 +35,31 @@ commands:
...
           at: /tmp/workspace
+      - load_image:
+          image_filename: << parameters.image_filename >>
+      - docker_login
+
+  # Load image from workspace
+  load_image:
+    description: "Load the image from the workspace"
+    parameters:
+      image_filename:
```
dpipemazo: :+1: nice

.circleci/atom.yml
```diff
@@ -54,45 +69,117 @@ commands:
         type: string
       service:
         type: string
+      image_tag:
```
dpipemazo: :+1: nice

.circleci/atom.yml
```diff
@@ -54,45 +69,117 @@ commands:
...
     steps:
       - run:
           name: Archive Docker image
-          command: docker save -o image.tar ${CIRCLE_PROJECT_REPONAME}-${CIRCLE_WORKFLOW_ID}
+          command: docker save -o << parameters.image_filename >>.tar << parameters.image_tag >>
       - persist_to_workspace:
           root: .
           paths:
-            - ./image.tar
+            - ./*.tar
```
dpipemazo: is there anything else here we're now persisting? Seems we went from a single file to a catch-all

.circleci/atom.yml
```diff
@@ -54,45 +69,117 @@ commands:
...
+      image_filename:
+        type: string
+      source_image:
+        type: string
+      target_image:
+        type: string
+    steps:
+      - load_image:
+          image_filename: << parameters.image_filename >>
+      - tag_and_deploy:
```
dpipemazo: maybe have different step names?

.circleci/config.yml
dpipemazo: now promoted to atom v0.0.2

.circleci/atom.yml
aywaldron: For some reason using `./<< parameters.image_filename >>.tar` consistently through a file not found error on circle, even though the file was clearly there. This works around that error and should only catch the single .tar file we want since no other .tar files are present in each job.

PR #245, opened 2020-01-31
Issue #236 - Parameterize docker push "no output timeout"

### LINKED ISSUE #236: Our custom orbs should allow parameterizing no_output_timeout
circleci has a default timeout of 10 minutes on run commands, which we have started hitting on some repos like defect-detection, causing us to fail our builds (see example failure here: https://app.circleci.com/jobs/github/elementary-robotics/element-defect-detection/175).
This timeout can be overwritten by adding `no_output_timeout`  to the command, which is in seconds. 

For example:
```
name: Generate local release (azure - with classifiers)
command: ansible-playbook -c local -u circleci -i hosts -l local-hlk-azure deploy.yaml --tags build --extra-vars "enable_advanced_tcp_tuning=false do_classification=true"
no_output_timeout: 1200
```

Unfortunately, because we use our custom orbs now in most of our containers, we can no longer set this directly, so `element-defect-detection` in particular at the moment is incapable of building on circle. We should either set this setting to something really high on all the custom orb commands, or better yet, find a way to expose this as a custom parameter from our orbs.

aywaldron:
Parameterized the `docker push` command's `no_output_timeout` parameter in atom orb, exposed through both the commands and jobs used to deploy images. This should be tested with dev orb ` elementaryrobotics/atom@dev:236` in `element-defect-detection` before publishing to prod and merging.

Resolves #236

### DIFF
--- a/.circleci/atom.yml
+++ b/.circleci/atom.yml
@@ -136,6 +136,10 @@ commands:
       target_tag:
         description: New tag to push
         type: string
+      output_timeout:
+        description: Elapsed time docker push can run without output
+        type: string  # decimal with unit suffix, such as “20m”, “1.25h”, “5s”
+        default: 15m
     steps:
       - run:
           name: Docker Tag
@@ -143,6 +147,7 @@ commands:
       - run:
           name: Docker Push
           command: docker push << parameters.target_image >>:<< parameters.target_tag >>
+          no_output_timeout: << parameters.output_timeout >>
 
   push_dev_image:
     parameters:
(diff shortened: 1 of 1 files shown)


aywaldron:
Successfully tested at https://github.com/elementary-robotics/element-defect-detection/pull/60. Orb should be promoted to prod.

.circleci/atom.yml
```diff
@@ -1,3 +1,14 @@
+#
+# Section for setting repeatedly used yaml anchors/aliases
+#
+aliases:
+  - &output_timeout_param
+      description: Elapsed time command can run without output
+      type: string  # decimal with unit suffix, such as “20m”, “1.25h”, “5s”
+      default: 15m
```
plipay: :+1:

aywaldron:
Orb promoted to prod version 0.0.4.

PR #249, opened 2020-02-05
Remove git hash from master tags (#248)

### LINKED ISSUE #248: Master build still tags with commit hash on docker hub
When building `master` from atom we push a tag `master-${CI_BUILD_NUM}-${COMMIT_HASH}` whereas we do not add the commit hash on the development builds. I think the commit hash can be removed to make it cleaner/easier to use the master tags.

aywaldron:
Resolves #248 

Published dev orb at `elementaryrobotics/atom@dev:243` which should be promoted to prod

### DIFF
--- a/.circleci/README.md
+++ b/.circleci/README.md
@@ -149,6 +149,19 @@ circleci orb publish promote elementaryrobotics/atom@dev:some-tag patch
 
 ### Release Notes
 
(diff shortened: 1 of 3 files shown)


dpipemazo:
@aywaldron `dev:243` published to prod and is now `v0.0.5`. Please update the release notes with this PR as well.

PR #252, opened 2020-02-06
Issue #250 - Rely on element for deserialization rather than CLI

### LINKED ISSUE #250: Atom CLI command/response records throw deserialization error
Atom CLI `records cmdres` needs to be updated to check whether a response was automatically deserialized by `entry_read_since`.

aywaldron:
Resolves #250. 

Possible similar issue noted at https://github.com/elementary-robotics/atom/blob/c707e6e6da60dfe63a906cdec42e3825c7a4f9d8/utilities/atom-cli/atom-cli.py#L244

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ aliases:
 version: 2.1
 
 orbs:
-  atom:  elementaryrobotics/atom@dev:251
+  atom:  elementaryrobotics/atom@0.0.6
 
 jobs:
   build-atom:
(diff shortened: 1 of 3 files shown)


aywaldron:
I think the fix in [this commit](https://github.com/elementary-robotics/atom/pull/252/commits/abe347201f6d7d14efd58982af8ae09d9fe7e49f) should be made but it is currently breaking several methods in atom that depend on it, so let's hold off on making this change right now. [This workaround](https://github.com/elementary-robotics/atom/pull/252/commits/ab91aaf2b1d3fa759cbf9b2fcf71a9087a627ae6) should fix the `cmdres` issue without causing other breaking changes for the time being.

PR #263, opened 2020-02-12
Improve type checking for pyarrow serialization: allow numpy arrays, …

### LINKED ISSUE #262: Allow arrow serialization of numpy arrays
Per @Cody-G, we want to allow `np.ndarray` types to be arrow serialized.

Additionally, want to make the type check here recursive so that it checks the types of elements of dicts, lists, etc.
https://github.com/elementary-robotics/atom/blob/bac43d7c36e72364e2aef3388367320b5fd78cb2/languages/python/atom/serialization.py#L46

Cody-G:
…don't allow builtins that result in pickling.

closes #262

### DIFF
--- a/languages/python/atom/serialization.py
+++ b/languages/python/atom/serialization.py
@@ -1,4 +1,5 @@
 from msgpack import packb, unpackb
+import numpy as np
 import pyarrow as pa
 from enum import Enum
 import builtins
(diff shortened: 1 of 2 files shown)


Cody-G:
Unfortunately I couldn't find any way to query for which types pyarrow will fall back to pickling.  I noticed that checking the builtins isn't quite sufficient; pyarrow will happily pickle an instance of `function` or `type`, so I added explicit checks for those.  It's a bit ugly, but the combination of those checks and pyarrow's default `SerializationContext` should prevent accidental pickling.

Cody-G:
(docs failure is unrelated)

Cody-G:
> Looks good to me -- a unit test for nested non-arrow-serializable type could be a good addition

Ah good idea, I just added a test for that.  If it passes I'll merge

PR #267, opened 2020-02-13
element: support force_serialization flag in entry APIs

dpipemazo:


languages/python/atom/element.py
```diff
@@ -973,11 +988,17 @@ def reference_get(self, *keys, serialization=None, deserialize=None):
...
             key_split = key.split(':') if type(key) == str else key.decode().split(':')
 
+            # Need to reformat the data into a dictionary with a "ser"
+            #   key like it comes in on entries to use the shared logic function
             if "ser" in key_split:
-                serialization = key_split[key_split.index("ser") + 1]
-            elif deserialize is not None:  # check for deprecated legacy mode
-                serialization = "msgpack" if deserialize else None
-
+                serialization = self._get_serialization_method(
```
aywaldron: Shouldn't this logic execute even if `"ser"` is not in the reference key?

dpipemazo: ah, yes. That's a good catch -- otherwise we'd be breaking the functionality on the legacy elements

languages/python/atom/element.py
```diff
@@ -973,11 +1000,22 @@ def reference_get(self, *keys, serialization=None, deserialize=None):
...
             key_split = key.split(':') if type(key) == str else key.decode().split(':')
 
+            # Need to reformat the data into a dictionary with a "ser"
+            #   key like it comes in on entries to use the shared logic function
+            get_serialization_data = {}
             if "ser" in key_split:
-                serialization = key_split[key_split.index("ser") + 1]
-            elif deserialize is not None:  # check for deprecated legacy mode
-                serialization = "msgpack" if deserialize else None
+                get_serialization_data["ser"] = key_split[key_split.index("ser") + 1]
```
aywaldron: :+1:

Issue #256, opened 2020-02-11
Rewrite atom and nucleus Dockerfiles for caching


aywaldron:
- Optimize DLC used in CircleCI
- Remove use of atom-base image
- Use multi-stage builds and buster/slim to decrease image size
- Distinguish between production and testing requirements so we aren't shipping our prod images bloated with test dependencies

aywaldron:
Closed by PR #270

PR #270, opened 2020-02-14
Issue #256 - Rewrite atom and nucleus Dockerfiles

aywaldron:
- Build both atom and nucleus from `buster`.
- Rearrange commands to maximize chances of docker layer caching in build.
- Ignore everything by default in docker context, specifying what not to ignore.
- Combine `Dockerfile-atom` and `Dockerfile-atom-base` into a single multistage `Dockerfile-atom` with `base`, `prod`, `test`, and `graphic` stages. The `prod` stage image does not contain build, test, or graphics dependencies. To build a specific stage, use
    ```
    docker build -t <tag> -f Dockerfile-atom --target=<base | prod | graphics | test> .
    ```
   Building a later stage will leave freshly built earlier stages dangling. Since the `prod` stage is not built from the `base` stage, changes in `base` will not disqualify `prod` from using DLC.

    Read more on multistage builds [here](https://docs.docker.com/develop/develop-images/multistage-build/).

- Make `nucleus` a multistage build with `base` and `prod` stages. The `prod` stage does not contain build dependencies. Since the `prod` stage is not built from the `base` stage, changes in `base` do not disqualify `prod` from using DLC.
- Uncompressed image sizes are currently:
    ```
     nucleus-base          919MB
     nucleus-prod          480MB  
     atom-base             819MB  
     atom-prod             471MB  (prev. 1.52GB - no longer includes graphics)
     atom-graphics         878MB  (prev 1.52GB - with graphics)
     atom-test             855MB
     atom-opengl           1.13GB  (prev. 1.71 GB - with graphics)
     atom-cuda-10          1.7GB  (prev. 3.25GB - no longer includes graphics)
     atom-opengl-cuda-10   2.9GB  (prev. 3.47GB - with graphics)
    ```

- Use a virtualenv to manage python dependencies. See [this article](https://pythonspeed.com/articles/multi-stage-docker-python/) for details.

**To do:**
- [ ] I'm not confident that the C/C++ atom libraries are being tested or copied over correctly - would like some feedback on this.
- [x] Needs VNC/noVNC graphics tested.
- [x] Build process needs updating for multi-stage builds.
- [x] Updates to any docs referencing the `atom-base` image


Dockerfile-atom
```diff
@@ -1,55 +1,167 @@
...
+# Install pytest
+RUN pip3 install --no-cache-dir pytest
+
+# Add code
+
+# Copy installed python libs
+COPY --from=base /root/.local /root/.local
+ENV PATH=/root/.local/bin:$PATH
+
+# Copy C builds
```
dpipemazo: Why not do the test build from `prod_final`?

aywaldron: Because I wanted to be able to use DLC on the test dependencies installs. The `COPY`s in `prod_final` would usually disqualify `test` from using DLC since they are likely to change.

Cody-G:
> atom uses local installs to manage python dependencies, while nucleus use a virtualenv. Interested to hear other's thoughts on this - I'm considering moving atom to use a virtualenv as well. See this article for details.

Nice article, thanks for the link.  It does seem compelling to use a virtualenv.  But I think this would require all elements to activate the virtualenv before they can use atom, correct?  Can we automate the activation of the virtualenv so that it doesn't break current element builds?

aywaldron:
@Cody-G Glad you enjoyed it! It actually links back to a [related article](https://pythonspeed.com/articles/activate-virtualenv-dockerfile/), which explains that the activation of a virtualenv can be achieved by adding its directory to the `PATH`, which is currently being done, for example here:
https://github.com/elementary-robotics/atom/blob/3ddba5c4fb05ba07acddc8aa18c1024afb164cbb/Dockerfile-nucleus#L51

Since environment variables set with `ENV` are passed onto images from their base images, this would be preserved in any subsequent image built from `atom`.

Dockerfile-atom
aywaldron: Updating this per our conversation

aywaldron:
The build process has now been updated to build each stage individually and push the production stages to dockerhub. The `atom` and `nucleus` images are now built in parallel before running the unit tests. The opengl/cuda images are now built from scratch on every build - this shouldn't add much time with the DLC.

The orb has been updated and published at `elementaryrobotics/atom@dev:256` - it should be promoted to prod before merging.

README.md
```diff
@@ -260,3 +260,9 @@ The combined OpenGL/cuda/cuDNN base image can be built in two steps, with the fi
 docker build --no-cache -f Dockerfile-install-cudnn -t opengl-cuda-10-base .
 docker build --no-cache -f Dockerfile-atom -t elementaryrobotics/atom-opengl-cuda-10-base:dev --target=graphics --build-arg BASE_IMAGE=opengl-cuda-10-base .
 ```
+
+A note on the use of the graphics included in the `opengl` images: The `DISPLAY` environment variable set to `:0`
+in the image is required for use of graphics through VNC. If you want to run graphics locally, you should forward your
+local `DISPLAY` variable to the element by specifying it without a value in the `environments` section of your docker-compose
+file. If you want to run graphics with VNC, you should either not include it in the docker-compose file so that the default
```
Cody-G: Correct `DISPLAY` configuration is necessary but not sufficient to use VNC graphics.  VNC usage is pretty complicated and I wonder if it's misleading to document only this piece here.  We have a VNC section in [Atom's graphics docs](https://github.com/elementary-robotics/atom/blob/master/doc/source/includes/sdk-graphics.md), maybe it's better to add the note there?  Apologies as I was the last person to work on those docs and I didn't include this because I wasn't aware of how DISPLAY interacted with the VNC.

Cody-G: Oh actually I did mention it in the last line of the linked page, but it's easy to miss because it's not in the VNC section.

aywaldron: Sure, I'll move this to that page of the docs.

dpipemazo:
Seems the only thing left on this is to test against the robot element for C/C++ builds -- perhaps @reidwagner can take a look to give sign-off. 

I do find using the `virtualenv` in a Docker image interesting. What are some of the benefits of doing this? I might have missed the article referenced in the high-level comments.

aywaldron:
Hi @dpipemazo, the use of a `virtualenv` makes it much easier to copy python installs from the `build` stage to a subsequent stage in multi-stage builds. As outlined in the article linked above:

> Python packages install many things in many places—if you just do a normal pip install, you will end up with:
>   - Code and other files (e.g. .pth files) installed in your Python install’s site-packages.
>   - Scripts installed in /usr/bin.
>   - Data files installed somewhere else again, usually under /usr.
>
> In the general case, figuring out all the files you need to copy from the compiler stage to the runtime stage can be difficult, and the resulting configuration will be brittle.

Using a `virtualenv` ensures that all required files are under the venv's directory, so only a single `COPY` command in subsequent docker stages is required.

reidwagner:
> Seems the only thing left on this is to test against the robot element for C/C++ builds -- perhaps @reidwagner can take a look to give sign-off.
> 
> I do find using the `virtualenv` in a Docker image interesting. What are some of the benefits of doing this? I might have missed the article referenced in the high-level comments.

I can definitely take a look.

dpipemazo:
@aywaldron we're going to fix a few other things in some of our C/C++ code to adapt to the new atom changes and it might take a bit longer -- let's get this PR in and we'll file a separate issue if C/C++ are broken. Since we have the new versioning it shouldn't be a big issue.

Issue #275, opened 2020-02-25
Tagged builds using our orb bump `latest`


dpipemazo:
`latest` should only be bumped from `master` builds -- tags should not also bump `latest` and I believe they currently are

aywaldron:
@dpipemazo This is actually a change that would need to be made to individual build config.ymls, not the orb. I pushed this change to atom and to the orb example config.

dpipemazo:
Ah, I see. In talking to the CircleCI team they'll be adding workflows to orbs in the future that would help us a bit with this.

PR #283, opened 2020-03-06
Remove latest bump from tagged builds (Issue #275)

aywaldron:
See #275 

An initial build on this PR failed because an `apt-get update` was cached, but the version of a library it tried to download was no longer available. I created a `LAST_UPDATED` env var within the Dockerfile to use for invalidating the `apt-get update` cache when necessary.

### DIFF
--- a/Dockerfile-atom
+++ b/Dockerfile-atom
@@ -68,6 +68,10 @@ WORKDIR /atom
 
 FROM $BASE_IMAGE as prod
 
+# Cache buster env var - change date to invalidate subsequent caching
+# See the atom README "Atom Dockerfile" section for more information
+ENV LAST_UPDATED 2019-03-06
+
 # Install python
 RUN apt-get update -y \
  && apt-get install -y --no-install-recommends apt-utils \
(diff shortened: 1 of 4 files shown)


Dockerfile-atom
```diff
@@ -68,6 +68,8 @@ WORKDIR /atom
 
 FROM $BASE_IMAGE as prod
 
+ENV LAST_UPDATED 2019-03-06
```
dpipemazo: Can you please detail a bit more about what this does and/or document it somewhere? Seems like a good fix but I can't find much with a quick google search

aywaldron: Sure. The variable itself is not used for anything and its name and value are effectively arbitrary - it just functions as a cache buster since it is before the `apt-get` installs, so changing it will prevent them from being cached. Using the last updated date just serves as a reminder for those using it/debugging. I will add note of this in the docs somewhere.

aywaldron: We could set multiple of these env vars at different points in the build if we wanted finer control over what is or is not cached, for example one in each of the build stages.

dpipemazo: ah, cool. Thanks for the explanation. Let's also add a comment in the dockerfile about it perhaps giving a pithy explanation and pointing the user towards the README for more info

dpipemazo:
:+1: looks good.

PR #284, opened 2020-03-06
Issue #276 - Add flake8 orb command

### LINKED ISSUE #276: Add command to orb to run standard `flake8` check against the code
We've generally agreed on internal Python coding style with the following `flake8` config settings:

```
[flake8]
max-line-length = 120
max-doc-length = 80
select = C,E,F,W,B
ignore = E203,W503
```

We should have a command in our orb that allows us to check for these. We don't need to apply it on all repos (and I'm not sure `atom/python` would pass for now), but it would help with other repos throughout the company.

aywaldron:
Resolves #276 

Orb needs promotion to prod.

### DIFF
--- a/.circleci/README.md
+++ b/.circleci/README.md
@@ -163,15 +163,14 @@ orbs:
   atom: elementaryrobotics/atom@0.0.8
 ```
 
-- Use the new command for installing and running `flake8`, optionally specifying which directory to run it in and which directories to exclude from the check:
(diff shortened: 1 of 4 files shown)


dpipemazo:
@aywaldron orb is now `0.0.8`

aywaldron:
@dpipemazo If we want to create a .flake8 file in each repository using it, we could do that instead of creating it in Circle. It might be easier to follow that way but would require repo users to create the file.

PR #285, opened 2020-03-10
Update flake8 check to run as job in docker image

aywaldron:
Update to PR #284 addressing Issue #276.

The flake8 format check has been changed from a command to a job, which is run in the `alpine/flake8` docker image. The python version to use is now a parameter which also determines which image tag to use. The available image tags corresponding to python versions can be viewed on [dockerhub](https://hub.docker.com/r/alpine/flake8). The default python version/image tag is set to `3.7.0`. See updated docs included in this PR for instructions on use.

This orb will need promotion to production version `0.0.9`.

### DIFF
--- a/.circleci/README.md
+++ b/.circleci/README.md
@@ -149,6 +149,51 @@ circleci orb publish promote elementaryrobotics/atom@dev:some-tag patch
 
 ### Release Notes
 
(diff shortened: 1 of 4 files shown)


dpipemazo:
@aywaldron I think this is good to merge :tada: If so, let's get it in :+1:

PR #287, opened 2020-03-14
286/add submodule orb

dpipemazo:
Dev orb published under `dev:286`. Figured it would be nice to share the submodule cloning command as a shared command in the orb. Eventually we might want a few separate orbs -- `atom` and `dev-basics` or something like that but for now having it all in `atom` is likely OK. 

One note is there is now a `--jobs` flag for the submodule update which I believe parallelizes the clone in order to speed it up a touch if there are multiple submodules being cloned. At a quick glance this seems like it perhaps took our "Update Submodules" step down from 9s to 6s.

### DIFF
--- a/.circleci/README.md
+++ b/.circleci/README.md
@@ -149,6 +149,40 @@ circleci orb publish promote elementaryrobotics/atom@dev:some-tag patch
 
 ### Release Notes
 
+#### [v0.0.10](https://circleci.com/orbs/registry/orb/elementaryrobotics/atom?version=0.0.10)
+Created 03/12/2020.
+
+##### New features
+- Added `update_submodules` command so that all elements with submodules don't have to write their own.
+- Parallelized `update_submodules` across 8 jobs s.t. it speeds up the clone a bit to save some build time/$$.
+
+##### Upgrade Steps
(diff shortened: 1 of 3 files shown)


.circleci/README.md
```diff
@@ -149,6 +149,40 @@ circleci orb publish promote elementaryrobotics/atom@dev:some-tag patch
...
+     executor: atom/build-classic
+     resource_class: large
+     environment:
+       <<: *atom_build_vars
+     steps:
+       - checkout
++      - atom/update_submodules
+       - set_atom_version
+       - atom/docker_login
+
```
aywaldron: Don't necessarily need to change this but generally I try to make the examples in the release notes non-atom specific so its easier for other folks to copy paste from when using in their builds. e.g.
```diff
 build:
     executor: atom/build-classic
     resource_class: large
     steps:
       - checkout
+      - atom/update_submodules
       - atom/docker_login
```

dpipemazo: ah, thanks. yeah, that makes sense. I don't think it's quite worth a re-build/push but it is a good thing to keep in mind. Thanks for pointing it out!

dpipemazo:
orb promoted to `v0.10.0`

PR #292, opened 2020-04-07
Dockerfile-atom: Set up LD_LIBRARY_PATH

dpipemazo:
This allows us to natively find the C/C++ atom libraries
in /usr/local/lib when compiling against them without having to
add the same line in this commit to each Dockerfile of an element
being built with the C/C++ bindings


dpipemazo:
This turns out not to be a good idea -- setting this variable globally here can mess with install paths elsewhere.

PR #319, opened 2020-05-11
301/opencv

dpipemazo:
- Adds in builds for `amd64` and `aarch64` with OpenCV `v4.3.0` installed
- Updates base image builds to chain subsequent builds + dockerfiles
- Adds new variants for `opencv`
- Any variant using VNC now has option with and without VNC.
- Updates docs.

### DIFF
--- a/Dockerfile-opencv
+++ b/Dockerfile-opencv
@@ -13,7 +13,13 @@ ARG DEBIAN_FRONTEND=noninteractive
 # Install pre-requisites
 RUN apt-get update && apt-get install -y \
     zlib1g-dev \
-    libjpeg62-turbo-dev
+    libjpeg62-turbo-dev \
+    libpng-dev \
+    libtiff-dev \
+    libopenexr-dev \
+    libavcodec-dev \
+    libavformat-dev \
+    libswscale-dev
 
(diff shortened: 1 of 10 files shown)


Dockerfile-opencv
```diff
@@ -0,0 +1,87 @@
...
+RUN diff --new-line-format="" --unchanged-line-format=""  /tmp/required_libs.txt /tmp/existing_libs.txt | grep -v /usr/local/lib > /tmp/libs_to_copy.txt
+RUN xargs -a /tmp/libs_to_copy.txt cp -L -t /usr/local/lib
+
+#
+# Clean up and only ship the following folders:
+#   1. /usr/local/lib
+#   2. /usr/local/include
+#   3. /opt/venv
+#
+FROM ${BASE_IMAGE} as opencv-base
```
aywaldron, line 83: Interesting, so this isn't actually built from the `production_image`, its just adding the libraries from it.

dpipemazo: right, it's built atop the previous base and just uses the production image to identify which libraries it needs to package up along with the opencv install to make sure everything works.

Issue #326, opened 2020-05-15
Build OpenCV with graphics


Cody-G:
It can be handy to have access to OpenCV's built-in functionality for displaying images.  With our current build OpenCV throws this error with instructions on how to build for display:

```
cv2.error: OpenCV(4.3.0) /atom/third-party/opencv/modules/highgui/src/window.cpp:651: error: (-2:Unspecified error) The function is not implemented. Rebuild the library with Windows, GTK+ 2.x or Cocoa support. If you are on Ubuntu or Debian, install libgtk2.0-dev and pkg-config, then re-run cmake or configure script in function 'cvShowImage'
```

This may make a big difference in image size.  If that's an issue this isn't a necessity.

PR #327, opened 2020-05-18
Docker: Overhaul build system

### LINKED ISSUE #297: VNC switch to 1080p resolution
VNC can be changed to be a more standard resolution, like 1080p, through editing the section of the `/etc/X11/xorg.conf` to be more like the following:

```
Section "Screen"
  Identifier "dummy_screen"
  Device "dummy_videocard"
  Monitor "dummy_monitor"
  DefaultDepth 24
  SubSection "Display"
    Viewport 0 0
    Depth 24
    Modes "1920x1080"
  EndSubSection
EndSection
```

dpipemazo:
**Overview**

This PR started out as a fairly simple desire to upgrade to Redis 6 #312 . Since Redis 6 is in the "stock" base that all of the OpenCV and Graphics get built atop, upgrading Redis 6 means essentially rebuilding all of the bases -- including for `aarch64`. These builds can take a long time, and in the case of some graphics bases are currently timing out since they can't be completed within 5 hours (CirceCI hard job limit) on a `large` VM on CircleCI. Unfortunately, upgrading to an `xlarge` VM on CircleCI is 10x the cost of a large for only 2x the cores and doesn't make sense financially. As such, this project turned into a bit of an exploration to determine how we could break up our monolithic base builds into the product of several different stages.

After going down the rabbit hole a bit it was determined that a simple build script that iterates over a directory of Dockerfiles could make sense, with the Dockerfiles ordered numerically to determine their order. We could pass arguments to these Dockerfiles using additional files in the directory structure. This is a pretty basic system and should eventually be replaced by something fancier but gets the job done for now.

Now, we could build images that are the result of several chained Dockerfiles without having to do too much more than put some Dockerfiles in a folder. While this is nice and simple, it yields quite large images as we will have at least one layer (and likely more) per Dockerfile. One thing that could help here was the work done in the OpenCV base in which we use [`ldd`](http://man7.org/linux/man-pages/man1/ldd.1.html) to look at a particular library/binary that we built in order to see what it depends on. We then copy the dependencies from wherever they are in the rootfs into `/usr/local/lib` and set `LD_LIBRARY_PATH=/usr/local/lib:${LD_LIBRARY_PATH}` so that the libs get picked up in their new location. Using a technique similar to this, we can:
1. Look at all binaries and libraries we built in `/usr/local/lib` and `/usr/local/bin` and determine their dependencies using `ldd`
2. Look at the libraries present in our final "production" image (typically `debian:buster-slim`).
3. Compare list (1) vs list (2) and copy the difference from the built Dockerfile and into `/usr/local/lib`

Then, we can package up the directories of the rootfs with our build product into a tarball at the end of our chained Dockerfile build including:
- `/usr/local/lib`
- `/usr/local/bin`
- `/usr/local/include` 
- `/opt/venv` and `/usr/lib/python3` for Python
- A few other minor things

Finally, we can copy this tarball out of the final container and create our "production" build fresh from our "production" container (`debian:buster-slim`), add in the tarball, install one or two packages and we get a three-ish-layer production image that has the full build product of our chain of Dockerfiles + only the libraries actually used by the things we care about. This creates fairly small images without the hassle of tracking down every `.so` manually and giving us decent flexibility.

This works fairly well for basic C/C++/Python builds/installs but gets a bit more complicated when it comes to some graphics-related things such as `PyQt5` and the VNC. As such, we have some additional arguments in the final packaging steps that can specify additional files in the rootfs to copy into the production image as well as additonal libraries/binaries to look at with `ldd` to make sure we get all dependencies. We can also specify packages to install in the final stage of the production image.

All of this is implemented in the new `docker` folder. This should (and likely will) turn into its own open-source project outside of Atom as it's generally useful when building containers with complex dependencies that need guaranteed performance/compatibility between ARM and AMD architectures. For now it will live here until it's split out. If we split it out into its own organization we can actually get 400,000 free build credits per month from CircleCI (around $300) which would be wonderful as the base image builds are typically what eats up a bunch of our compute credits.

**Collateral Work**

While working on the above, it was determined that the `opengl` and `vnc` variants are pretty broken in master. These have been fixed and the VNC now works. The VNC did undergo a few structural changes in order to modernize it and remove the dependencies on `python2`. As such, the code has been internalized into the `utilities` folder and no longer has a dependency on `docker-opengl`. In doing so, the VNC was updated to use 1080p resolution which fixes #297 .

This also adds Qt5 + PyQt5 to the `graphics` build.

Finally, we should be able to build `extras` on top of `graphics` to solve #326 .

**Remaining Work**

- [x] Implement new build system
- [ ] Determine which "variants" we want to ship
- [ ] Update CircleCI to use new system
- [ ] Rebuild and test new base images on CircleCI


dpipemazo:
Closing this as we now have ARM images in CircleCI so we need to play less games/tricks to get this to work.

Issue #331, opened 2020-05-22
Parameterize Element Creation timeout


plipay:
Currently, if you try to create an Element instance to connect to some remote host, but the remote host is down, we block for an absurdly long time before raising an Exception. You can test this locally by doing something like:
```
from atom import Element
a = Element('test', host='banana')
```

Most likely, this is happening somewhere in the redis client constructor.
What we would like, is to have this timeout be parameterizable.
Find out what the default is today, and maintain that, but allow us to do something like:
```
a = Element('test', host='banana', timeout_ms=3000)
```
And have the element throw an exception after 3 seconds.

nceruchalu:
This will be handy in the heartbeat logic. We currently do a bunch of "interesting" things to work around this.

mvanveen:
@plipay any preferences on a default timeout in this context?

mvanveen:
:ship: 'd

PR #332, opened 2020-05-22
introduce a conn_timeout_ms kwarg to Element

mvanveen:
Implements #331


languages/python/atom/config.py
```diff
@@ -22,3 +22,5 @@
 ATOM_CALLBACK_FAILED = 7
 ATOM_LANGUAGE_ERRORS_BEGIN = 100
 ATOM_USER_ERRORS_BEGIN = 1000
+
+VERSION = "1.5.0-4-gfea183a-dirty"
```
mvanveen, line 26: @plipay I think the build system injects this `VERSION` into my branch for me, right?  I assume there's probably something manual I need to here to fit into our versioning schema.  Thoughts/advice?

dpipemazo: Please don't check this into version control -- it gets auto-generated by the build scripts.

dpipemazo: The reason this needs to exist is that we don't have the full git repo in the Docker container when we're building.

Issue #167, opened 2019-06-25
Fix bug with how we are using xread by using xreadgroup instead


plipay:
Right now, our xread calls in command_send are actually pulling the entire history of messages to that stream, because we are relying on self.response_last_id, which never gets updated.

We should actually move to using xreadgroup, which is a better fit for what we are trying to do here.

mvanveen:
### API Changes

**Proposed Process Architecture**

Currently `command_loop` often lives as a separate thread within a parent process.

For example, in `spec-robot` a `CommandThread` is instantiated as an attribute 
of `SpecRobot` and subsequently `start`ed by the parent thread.

The above proposed API change would not change the above portrait at all, but
would instead extend it to fork `n_procs` worth of worker processes.  `n_procs`
would be set `1` by default in the trivial case:

![image](https://user-images.githubusercontent.com/221196/82388268-10b84780-99ee-11ea-8e12-6f1c0569c394.png)

**Note that in the above scenario, while the parent `command_loop` routine lives in a `CommandThread` it could alternatively live in a separate process after this PR.**

`Element` objects typically behave as singletons AFAIK so multi-process coordination

dpipemazo:
@mvanveen this looks great! One thing I noticed:

> the parent command_loop routine would establish a consumer group name for the workers to share, as well as forking out child worker processes

I think this is close but perhaps not quite the desired behavior. It would be quite interesting if the same code, with no modifications, can work whether we spin up 3x copies of an element container with `n_procs=1` or spin up 1 copy of an element container with `n_procs=3`. To do this, I think that there needs to be a globally implied/specified consumer group name for reading from command streams that **all** elements with the same name use. What do you think?

dpipemazo:
Also, if we do the above -- I think the multi-processing problem actually became a bunch easier. When we make the children, as long as they have an element **with the same name** as the parent, then we can just so a simple fork/copy and there's no shared memory required. The thread that called `command_loop` would retain the original element, and a deep copy of the element would be passed to each child in `command_loop`. The user of the element needs to ensure that the rest of the app/handlers are process-safe but I think this will simplify it significantly and actually come out quite cleanly.

dpipemazo:
In terms of creating the consumer group -- looking at the `XGROUP` docs we can actually just have each individual worker call `XGROUP CREATE` as redis handles if the group already exists cleanly, throws a discernable error, and we can go on our merry command-handling way.

https://redis.io/commands/xgroup

[...]

PR #336, opened 2020-05-28
#167 parallel multi-process command_loop execution

mvanveen:
Implements #167 .

- [X] ~~**TODO:** would still benefit from end-to-end integration testing with other components~~
  - _update: has now been tested with the defect detection routine on the ir01local sim and n_procs > 3_

## Changes

In addition to the changes discussed in #167, there were a few subsequent noteworthy 
code modifications in this PR:

- In addition to `n_procs`, `command_loop` has two other new keyword arguments:
  - `block`: whether or not `command_loop` should exhibit blocking execution to wait
                 until children worker processes are shutdown (defaulted to `True`)
  - `read_block_ms`: number of milliseconds to block for during `XREADGROUP` in main
                                 command_loop event loop (defaulted to 1000 ms)

- the `element_name` argument of `get_all_commands` can now accept a list of element names instead of just a string.  If a string is provided the implementation is backwards compatible with previous semantics.
- command execution within a command loop resulting in an exception will now trigger an `ATOM_INTERNAL_ERROR` response and log the exception at `ERR` level.
  - this should log exceptions occurring during command execution within children `command_loop` processes to stdout in many contexts

## Notes

-  No stream consumer worker "reference counting" strategy was necessary for stream clean up (as was previously discussed).  The `NOGROUP` error responses previously encountered were ultimately a consequence of a bug where children worker processes didn't pick up the shutdown `multiprocessing.Event` properly.

- Semantics are such that the parent `Element` process (either the process which invokes `command_loop` or, alternatively, the process which constructed the `Element`) is the only pid that can clean up an `Element`'s streams
  - attempts to `__del__` in children process are currently explicitly logged at `ERR` level (testing has surfaced that this should be infrequent/exceptional under normal usage)
- each `command_loop` currently opens up its own redis connection to avoid issues with stale socket descriptors in the forked children process

- has been demonstrably tested with `command_loop` executed in the same process, from a separate `multiprocessing.Process` and `threading` in unit tests

**external process**
```python
>>> p = Process(target=element.command_loop, kwargs={'n_procs':2})
>>> p.start()
>>> # ...some time passes
>>> element.command_loop_shutdown()
>>> p.join()
```

**external thread**
```python
>>> t = Thread(target=element.command_loop, kwargs={'n_procs':2})
>>> t.start()
>>> # ...some time passes
>>> element.command_loop_shutdown()
>>> t.join()
```

**same process**
```python
>>> element.command_loop(n_procs=2, block=False)
>>> 
>>> element.command_loop_shutdown()
```
- for execution in same process you can call `command_loop` with `block=False` to fork out worker processes without blocking execution.  By default the method will block until the worker processes have received a shutdown event


dpipemazo:
> No stream consumer worker "reference counting" strategy was necessary for stream clean up (as was previously discussed). The NOGROUP error responses previously encountered were ultimately a consequence of a bug where children worker processes didn't pick up the shutdown multiprocessing.Event properly.

While there might not be errors -- I still think we want the reference counting behavior. Wouldn't the following still happen?:
1. N workers spin up, command/response streams are all created
2. A worker spins down, it cleans up the queues. This deletes the previous command/response queues
3. Additional commands/responses come in. When they come in they all work "fine" as `XADD` will create a stream, but we (a) lost all of the previous history and (b) could have a race/issue where someone blocks on a stream that doesn't exist for some amount of time?

languages/python/atom/element.py
```diff
@@ -137,21 +157,35 @@ def clean_up_stream(self, stream):
...
     def __del__(self):
-        """
-        Removes all elements with the same name.
-        """
+        """Removes all elements with the same name."""
+        # if a command loop is runnning, shut it down
+        self.command_loop_shutdown()
+        cur_pid = os.getpid()
+        if cur_pid != self._pid:
+            # we are in a child process pid. The parent pid has the
```
dpipemazo, line 173: I don't think this logic holds for the variant of this where we spin up multiple copies of the same element in separate Docker containers? I think it's a bit cleaner/more global if we use the refcount solution as discussed.

languages/python/atom/element.py
```diff
@@ -137,21 +157,35 @@ def clean_up_stream(self, stream):
...
+        self.command_loop_shutdown()
+        cur_pid = os.getpid()
+        if cur_pid != self._pid:
+            # we are in a child process pid. The parent pid has the 
+            # responsibility to clean up streams. This simplifies
+            # the stream collection model and avoids weird clean up 
+            # interleaves and race conditions
+            self.log(
+                LogLevel.ERR,
+                "Cowardly refusing to clean up streams during Element deletion"
```
dpipemazo, line 179: While this is witty, please keep logs strictly informational. We don't want to have discussions/judgement calls about phrasing/tone of voice in logs.

languages/python/atom/element.py
```diff
@@ -207,6 +241,15 @@ def _make_command_id(self, element_name):
...
         return f"command:{element_name}"
 
+    def _make_consumer_group_id(self, element_name):
+        """
+        Creates the string representation for an element's command stream id.
+
+        Args:
+            element_name (str): Name of the element to generate the id for.
+        """
+        return f"command:consumer_group:{element_name}"
```
dpipemazo, line 251: how is this different from the command stream?

languages/python/atom/element.py
```diff
@@ -420,8 +470,12 @@ def command_add(self, name, handler, timeout=RESPONSE_TIMEOUT, serialization=Non
         """
         if not callable(handler):
             raise TypeError("Passed in handler is not a function!")
-        if name in self.reserved_commands and name in self.handler_map:
-            raise ValueError(f"'{name}' is a reserved command name dedicated to {name} commands, choose another name")
+        if ((name in RESERVED_COMMANDS and name in self.handler_map)
```
dpipemazo, line 473: what's the difference between `RESERVED_COMMANDS` and `DISALLOWED_COMMANDS`?

languages/python/atom/element.py
```diff
@@ -483,22 +537,110 @@ def wait_for_elements_healthy(self, element_list, retry_interval=HEALTHCHECK_RET
...
+                                    from the function.
+                                       block.
+            read_block_ms (integer, optional): Number of milliseconds to block
+                                               for during a stream read insde of 
+                                               a command loop.
+        """
+        # update self._pid in case e.g. we were constructed in a parent thread but 
+        # `command_loop` was explicitly called as a sub-process
+        self._pid = os.getpid()
+        n_procs = int(n_procs)
```
dpipemazo, line 563: is this necessary?

languages/python/atom/element.py
```diff
@@ -483,22 +537,110 @@ def wait_for_elements_healthy(self, element_list, retry_interval=HEALTHCHECK_RET
...
+        assert n_procs > 0, "n_procs must be a positive integer"
```
dpipemazo, line 564: Is it standard to use `assert` here as opposed to some other sort of error handling?

languages/python/atom/element.py
```diff
@@ -483,22 +537,110 @@ def wait_for_elements_healthy(self, element_list, retry_interval=HEALTHCHECK_RET
...
+
+        children = []
+        parent = True
+        for i in range(n_procs):
+            child = os.fork()
```
dpipemazo, line 569: why are we using `fork` and not `multiprocessing`?

languages/python/atom/element.py
```diff
@@ -483,22 +537,110 @@ def wait_for_elements_healthy(self, element_list, retry_interval=HEALTHCHECK_RET
...
+                os.waitpid(cpid, 0)
+
+    def _command_loop(self, shutdown_event, read_block_ms=1000):
+        if hasattr(self, '_host'):
+            _rclient = redis.StrictRedis(host=self._host, port=self._port)
+        else:
+            _rclient = redis.StrictRedis(unix_socket_path=self._socket_path)
+        _pipe = _rclient.pipeline()
+        # get a group handle
+        # XXX: if use_command_last_id is set then the group will receive
```
dpipemazo, line 588: what's the `XXX` here?

languages/python/atom/element.py
```diff
@@ -483,22 +537,110 @@ def wait_for_elements_healthy(self, element_list, retry_interval=HEALTHCHECK_RET
...
+        group_name = self._make_consumer_group_id(self.name)
+        group_last_cmd_id = self.command_last_id 
+        try:
+            _rclient.xgroup_create(
+                stream_name,
+                group_name,
+                group_last_cmd_id,
+                mkstream=True
+            )
+        except redis.exceptions.ResponseError:
```
dpipemazo, line 603: Maybe a log here would be nice?

languages/python/atom/element.py
```diff
@@ -483,22 +537,110 @@ def wait_for_elements_healthy(self, element_list, retry_interval=HEALTHCHECK_RET
...
+            # If we encounter a `ResponseError` we assume it's because of a `BUSYGROUP`
+            # signal, implying the consumer group already exists for this command.
+            #
+            # Thus, we go on our merry way as we can successfully proceed pulling from the 
+            # already created group :)
+            pass
+        # make a new uuid for the consumer name
+        consumer_uuid = str(uuid.uuid4())
```
dpipemazo, line 611: should we double-check for uniqueness of the `uuid`? Unlikely to have clashes, but would be a weird bug. I think we can query for all clients in a group at the given time?

languages/python/atom/element.py
```diff
@@ -483,22 +537,110 @@ def wait_for_elements_healthy(self, element_list, retry_interval=HEALTHCHECK_RET
...
+        while not shutdown_event.is_set():
             # Get oldest new command from element's command stream
-            stream = {self._make_command_id(self.name): self.command_last_id}
-            cmd_responses = self._rclient.xread(stream, block=MAX_BLOCK, count=1)
+            # XXX: note: consumer group consumer id is implicitly announced
```
dpipemazo, line 614: `XXX` again

languages/python/atom/element.py
```diff
@@ -483,22 +537,110 @@ def wait_for_elements_healthy(self, element_list, retry_interval=HEALTHCHECK_RET
...
+                    group_name,
+                    consumer_uuid,
+                    {stream_name: '>'},
+                    block=read_block_ms,
+                    count=1
+                )
+            except redis.exceptions.ResponseError:
+                self.log(
+                    LogLevel.ERR,
+                    f"Attempted XREADGROUP on closed stream {stream_name} "
```
dpipemazo, line 626: this seems to assume a particular error. Is this error always the one thrown?

languages/python/atom/element.py
```diff
@@ -483,22 +537,110 @@ def wait_for_elements_healthy(self, element_list, retry_interval=HEALTHCHECK_RET
...
+                    _pipe=_pipe
+                )
+                return
+
             if not cmd_responses:
                 continue
-            stream_name, msgs = cmd_responses[0]
+            cmd_stream_name, msgs = cmd_responses[0]
+            assert cmd_stream_name.decode() == stream_name, \
+                    "Expected received stream name to match: %s %s" % (cmd_stream_name, stream_name)
```
dpipemazo, line 636: this indent seems a bit arbitrary -- I don't think we use 8 spaces anywhere.

languages/python/atom/element.py
```diff
@@ -519,46 +661,79 @@ def command_loop(self):
...
+                    except:
+                        self.log(
+                            LogLevel.ERR,
+                            "encountered error with command: %s\n%s" % (
+                                cmd_name,
+                                format_exc()
+                            ),
+                            _pipe=_pipe
+                        )
+                        # TODO: consider enabling debug mode with exception printing
```
dpipemazo, line 694: if we have a TODO in the comments, please ensure there's an issue filed on it.

languages/python/atom/element.py
```diff
@@ -519,46 +661,79 @@ def command_loop(self):
...
-            _pipe = self._release_pipeline(_pipe)
 
-    # Triggers graceful exit of command loop
+            # `XACK` the command we have just completed back to the consumer 
+            # group to remove the command from the consumer group pending 
+            # entry list (PEL).
+            _pipe.xack(
+                stream_name,
+                group_name, 
+                cmd_id #TODO bytes or int?
```
dpipemazo, line 730: `TODO` in comment.

languages/python/atom/element.py
```diff
@@ -519,46 +661,79 @@ def command_loop(self):
...
             kv["cmd"] = cmd_name
             _pipe.xadd(self._make_response_id(caller), kv, maxlen=STREAM_LEN)
             _pipe.execute()
-            _pipe = self._release_pipeline(_pipe)
 
-    # Triggers graceful exit of command loop
+            # `XACK` the command we have just completed back to the consumer 
+            # group to remove the command from the consumer group pending 
+            # entry list (PEL).
+            _pipe.xack(
```
dpipemazo, line 727: we should probably handle/log this failing.

languages/python/atom/element.py
mvanveen: How would this interact with the Client/Server design?  My understanding is the principle we were espousing was one process per element name, and the proposed use case with multiple docker containers seems like it would conflict.  I was under the auspice that each docker container would be logically a different Element.

Could we opt to have each docker container prescribe a unique Element name, or is that too burdensome?  I would like to understand your thoughts with respect to the above a little bit more.

Happy to drive the refcount solution but definitely want to understand the end-to-end ramifications such as the aforementioned docker scenario a bit better before I drive an implementation.  Thx!

dpipemazo: All good things to think through, however I think we want container-based scaling as a core feature in Atom and therefore need to make it work. Imagine running in something like kubernetes/ECS which can do load balancing.

dpipemazo: For the servers, they could all have the same element name, and they all have unique consumer IDs so I think it's all pretty much the same?

mvanveen:
Thank you for the review comments @dpipemazo 

I am definitely happy to look at #1 more closely.  I left a more direct comment I think is likely worth some discussion to make sure I'm on the same page as you.


With respect to #2 there are some pitfalls with `multiprocessing` that I wanted to avoid by simplifying with `os.fork` logic to guarantee that we truly were exhibiting the behaviors of a process fork.

Check out discussions of `spawn` vs `fork` vs `forkserver` models of `multiprocessing`, for example [this blog post highlighting an common deadlock scenario](https://pythonspeed.com/articles/python-multiprocessing/).  These kinds of semantic nuances made me a little cautious to jump right into `multiprocessing` primitives within `command_loop`.

That said, I have successfully tested an alternative implementation of `command_loop` that uses `multiprocessing.Process` successfully against the unit tests we have in the PR currently.

I'm still a little concerned/wary that I can't guarantee beyond a shadow of a doubt (especially without some careful context management and thorough testing) that moving to `multiprocessing` might incur some weird unintended side effects or dead lock potential, but given that I've successfully tested the above scenario with the existing tests I think we could probably move to `multiprocessing.Process` in `command_loop` if you deem it more idiomatically Pythonic.

languages/python/atom/element.py
mvanveen: good call, I'll add exception handling

languages/python/atom/element.py
mvanveen: likely an artifact of my minimal/non-existant `.vimrc` :man_facepalming: .  nice catch thx

languages/python/atom/element.py
mvanveen: it's kind of most closely related to my debugging hysteresis at this point.  Initially this seemed like a critical region trouble spot, in that it appeared that there was opportunity to hit the `XREADGROUP` call even after the command loop was shutdown.

After more thorough inspection it's become clear that this was because the `multiprocesing.Event` wasn't being properly picked up in past commits during development.

I decided to keep it intact in the PR in case we hit this case in the real world, as I figured it would be a valuable thing to log (perhaps similar in spirit to your comment with respect to the `XACK`) but I am more bearish on the need for this handling than I was previously.

languages/python/atom/element.py
mvanveen: I like explicitly casting to `int` to verify that an input satisfies a given duck type.  Not particularly tied to this line but personally like to err on defensiveness where possible.

languages/python/atom/element.py
mvanveen: It is not different :grinning: .  I realized late in implementation that it could conflict with a command group called `command_group`, and decided to disallow this use case.

A possibly more elegant strategy would be to just pick a different key space.  I've provided unit testing through this particular case with the existing semantics because having a command called `command_group` seems confusing to me, but it would not be difficult to merely pick a different keyspace for a command's respective command group that does not overlap with the command stream namespace whatsoever.

languages/python/atom/element.py
mvanveen: this is best situated with the answer in the above comment about the key generation for command groups, let's center discussion up there (there is some explanation already provided).

languages/python/atom/element.py
mvanveen: :+1: Roger, thanks for the feedback.  I felt it acceptable because it is a nod to [`tar`'s output](https://askubuntu.com/questions/617376/why-i-get-cowardly-refusing-to-create-an-empty-archive-with-tar-command).  Will stick to a more neutral voice in my error messages from here on out.

languages/python/atom/element.py
mvanveen: this is a tip I picked up from a coworker several jobs ago.  Apparently it is meant to [denote a heads up](https://stackoverflow.com/questions/1452934/what-is-the-meaning-of-xxx-in-code-comments) to readers of the source.

Some text editors will highlight this by default to ensure that future travelers can find it.

Happy to de-indent

languages/python/atom/element.py
mvanveen: FWIW this is a rather common occurrence.  Anything process except the first to the finish will encounter this exception.  Do we still feel it's appropriate to log in that circumstance, and what severity would we prefer?

Particularly because logs look like they are implemented as a ring buffer currently I would be concerned about the potential to swamp our logging with this scenario.

languages/python/atom/element.py
mvanveen: My understanding is that "if every person on the planet generates a new UUID4 every second, we’d expect a collision to occur after about 10 years." [source](https://www.quora.com/Has-there-ever-been-a-UUID-collision).

That said I'm a big fan of looking before you leap in code.  I believe `XINFO` can accommodate this use case.  I can add some handling for this case to be extra sure ;)

languages/python/atom/element.py
mvanveen: Note: The specific response I'm about to provide conflicts with my design in #337 somewhat, so want to call this out in case it creates confusion :wink: 

One interesting consequence of the above scenario is that this should essentially work as-is right now if I'm thinking through this properly (certainly we can do some quick integration testing to proof of concept and validate)-

- each docker container will have a "parent" element of a particular `pid` denoted by `os.getpid`.
  - within each docker container, each parent `element` subequently either forks, multi-threads, or calls a `command_loop` invocation in-process.
- `command_loop` within each container will currently fork workers.
  - each worker will generate a unique consumer group ID and start pulling tasks, essentially identical to if they had been forked out by one container instead of multiple.

Thoughts?

languages/python/atom/element.py
mvanveen: `¯\_(ツ)_/¯` I've never found a one-size-fits-all rule of thumb across the projects/communities/teams I've sampled on this.

In this specific case I was pretty conflicted about `assert` versus checking and raising e.g. `ValueError`.  I personally tend towards the latter and like using `assert`s more for exceptional "this is a thing I hold to be true" assumption than an aspect of behavior I'm specifically guarding against, if that makes sense.

Would love to have a more cohesive rule of thumb to follow with respect to our coding guidelines.  If I missed this in our coding style guide apologies for the oversight.

languages/python/atom/element.py
mvanveen: left some discussion on this below.  happy to talk through more!

languages/python/atom/element.py
mvanveen: re-read your inquiry and realized I misunderstood what you are asking here.  Hopefully the above answer at least provides some situational context.

I suspect there _could_ be more exceptional failure modes than this specific one if there are other `redis.exceptions.ResponseError`- I did feel emboldened noticing that we make this assumption in 
other places of the codebase (e.g. a `ResponseError` implies a connection wasn't achieved).

I'll revisit and choose a more neutral log message that doesn't indicate a specific failure.

languages/python/atom/element.py
mvanveen: Thinking through the above more, I think I catch your meaning better-

Specifically, while the above works for construction of command groups, during `__del__` if we're not careful we'll pre-emptively clean up streams that are still in use.

The proposed refcount strategy should probably get around this deficiency :+1: , but thankfully I think we could preserve the pic-checking sematics.

mvanveen: We will still need to revisit #337 I think at this point.  I'm not sure what guarantees we can put on `ServerElement` objects if we allow multiple instantiations of the same `Element` name.

dpipemazo:
> Check out discussions of spawn vs fork vs forkserver models of multiprocessing, for example this blog post highlighting an common deadlock scenario. These kinds of semantic nuances made me a little cautious to jump right into multiprocessing primitives within command_loop.

This comment makes sense, but we need to also consider that in the context of our processes we're:
1. Not using `pool` -- we're just spinning processes
2. Not sharing any memory -- each one is completely independent
3. Not using any additional threads

By design we can avoid a lot of the deadlock and IMO deadlock typically isn't to blame based on the way a process was spun up but instead on the design of the program.

languages/python/atom/element.py
```diff
@@ -483,22 +559,144 @@ def wait_for_elements_healthy(self, element_list, retry_interval=HEALTHCHECK_RET
...
+                                               a command loop.
+        """
+        # update self._pid in case e.g. we were constructed in a parent thread but 
+        # `command_loop` was explicitly called as a sub-process
+        self._pid = os.getpid()
+        n_procs = int(n_procs)
+        assert n_procs > 0, "n_procs must be a positive integer"
+
+        self.processes = []
+        for i in range(n_procs):
```
Cody-G, line 589: Would it be worth adding a check for the number of currently active Threads, and then throwing a warning (or maybe even an error) if there's more than one active Thread?  My understanding is that this could help us identify odd bugs that come from `fork`ing a multi-threaded process (since the threads won't get copied).  Michael I think you already brought this up so maybe it's solved in some other way already, just wondering what you think about doing a check here.

mvanveen: Hey Cody,

Good call out.  I think we can easily add a debug level log line around if the [number of active threads](https://stackoverflow.com/questions/12476888/how-to-check-the-number-of-currently-running-threads-in-python) is greater than one.

IIRC the thing I was thinking about previously was whether or not a child thread would be able to perform a fork out when `command_loop` opens a subsequent process.

I have specifically tested a scenario similar to the above in unit testing where I open up a `Thread` that invokes `command_loop` and found it worked perfectly well.   That said, the test harness did not explicitly require thread state from the parent process to be copied over and that is very much still an issue.    However, in practice I think we could and arguably should easily avoid this situation for running `command_loop`s.

Agree that the debug line could be useful in the sorts of settings you're describing.

Cody-G: Cool, thanks, that all makes sense.  My feeling is that with your work we'll have enough avenues to implement whatever form of parallelism we need, I just want to have warnings or docs for any avenues that aren't guaranteed to work.  Sounds like we're on the same page :-)

mvanveen: There's definitely a balance to be struck within these changes between giving folks enough rope to get tangled up versus not being flexible enough :/.  We had similar trade-off discussions around using `os.fork` versus `multiprocessing.Process` and settled on the latter to err more on the latter such that we can fail fast.

I agree and think it's similar here, but possibly erring on the other side.  A quick call out to multiple active threads will only help alert ppl to possible issues while not hopefully getting in their way too much if they're trying something kind of fancy or outside well-tested paths.

languages/python/atom/config.py
```diff
@@ -22,3 +22,5 @@
 ATOM_CALLBACK_FAILED = 7
 ATOM_LANGUAGE_ERRORS_BEGIN = 100
 ATOM_USER_ERRORS_BEGIN = 1000
+
+VERSION = "1.5.0-25-g237ccfc-dirty"
```
dpipemazo, line 26: please remove this

languages/python/atom/element.py
```diff
@@ -38,21 +51,24 @@ def __init__(self, name, host=None, port=DEFAULT_REDIS_PORT,
         self.timeouts = {}
         self._redis_connection_timeout = float(conn_timeout_ms / 1000.)
         assert self._redis_connection_timeout > 0, \
-                "timeout must be positive and non-zero"
+            "timeout must be positive and non-zero"
         self.streams = set()
         self._rclient = None
-        self._command_loop_shutdown = threading.Event()
+        self._command_loop_shutdown = multiprocessing.Event()
```
dpipemazo, line 57: :+1:

languages/python/atom/element.py
```diff
@@ -483,22 +559,153 @@ def wait_for_elements_healthy(self, element_list, retry_interval=HEALTHCHECK_RET
...
+                                       block.
+            read_block_ms (integer, optional): Number of milliseconds to block
+                                               for during a stream read insde of 
+                                               a command loop.
+        """
+        # update self._pid in case e.g. we were constructed in a parent thread but 
+        # `command_loop` was explicitly called as a sub-process
+        self._pid = os.getpid()
+        n_procs = int(n_procs)
+        assert n_procs > 0, "n_procs must be a positive integer"
```
dpipemazo, line 586: Please only use `assert` on truly fatal system-level checks (out of memory, etc.). If it's something the user can mess up, I think it should be an exception instead of an assertion.

languages/python/atom/element.py
```diff
@@ -483,22 +559,153 @@ def wait_for_elements_healthy(self, element_list, retry_interval=HEALTHCHECK_RET
...
+
+        thread_count = threading.active_count()
+        if thread_count > 1:
```
dpipemazo, line 589: Please add a comment describing the situation that can lead to this/how it should be done else-wise. I think it's a good log, but I imagine someone will come back to the source to figure out the issue?

languages/python/atom/element.py
```diff
@@ -483,22 +559,153 @@ def wait_for_elements_healthy(self, element_list, retry_interval=HEALTHCHECK_RET
...
+        for i in range(n_procs):
+            p = Process(target=self._command_loop, args=(self._command_loop_shutdown,))
+            p.start()
+            self.processes.append(p)
+
+        if block:
+            for p in self.processes:
+                p.join()
+
+    def _increment_command_group_counter(self, _pipe):
```
dpipemazo, line 607: Please add a docstring to all functions.

languages/python/atom/element.py
```diff
@@ -483,22 +559,153 @@ def wait_for_elements_healthy(self, element_list, retry_interval=HEALTHCHECK_RET
...
             # Get oldest new command from element's command stream
-            stream = {self._make_command_id(self.name): self.command_last_id}
-            cmd_responses = self._rclient.xread(stream, block=MAX_BLOCK, count=1)
+            # note: consumer group consumer id is implicitly announced
+            try:
+                cmd_responses = _rclient.xreadgroup(
+                    group_name,
+                    consumer_uuid,
+                    {stream_name: '>'},
+                    block=read_block_ms,
```
dpipemazo, line 680: I think this will time out if we don't get a command within 1s by default? I think that's a bit extreme and not in our typical usage pattern. I think the timeout should default be infinite?There are many command loops which sit untouched for a while in the system.

dpipemazo: I see this is perhaps caught in the `continue` -- so by design we're resetting the connection each second. I think this may be a bit much, 10s would perhaps be better?

languages/python/atom/element.py
```diff
@@ -483,22 +559,153 @@ def wait_for_elements_healthy(self, element_list, retry_interval=HEALTHCHECK_RET
...
+                    ),
+                    _pipe=_pipe
+                )
+                return
+
             if not cmd_responses:
                 continue
-            stream_name, msgs = cmd_responses[0]
+            cmd_stream_name, msgs = cmd_responses[0]
+            assert cmd_stream_name.decode() == stream_name, \
```
dpipemazo, line 700: Let's throw errors instead of `assert`ing

languages/python/atom/element.py
```diff
@@ -519,46 +726,89 @@ def command_loop(self):
...
+                        err_code=ATOM_CALLBACK_FAILED,
+                        err_str=f"Return type of {cmd_name} is not of type Response"
+                    )
 
+            # send response on appropriate stream
             kv = vars(response)
             kv["cmd_id"] = cmd_id
             kv["element"] = self.name
             kv["cmd"] = cmd_name
             _pipe.xadd(self._make_response_id(caller), kv, maxlen=STREAM_LEN)
```
dpipemazo, line 784: should we add exception handling around this as well? Seems that if we fail to XADD back the XACK won't get sent which will lead to the behavior that we keep retrying to send the command? This makes sense, just wanted to bring it up.

languages/python/atom/config.py
mvanveen: d'oh thanks!

languages/python/atom/element.py
mvanveen: sure I like that rule of thumb thanks for the guidance on style

languages/python/atom/element.py
mvanveen: oh even private members!  no problem.

languages/python/atom/element.py
mvanveen: sure thing

languages/python/atom/element.py
```diff
@@ -583,8 +583,18 @@ def command_loop(self, n_procs=1, block=True, read_block_ms=1000):
         # `command_loop` was explicitly called as a sub-process
         self._pid = os.getpid()
         n_procs = int(n_procs)
-        assert n_procs > 0, "n_procs must be a positive integer"
-
+        if n_procs <= 0:
+            raise ValueError("n_procs must be a positive integer")
+
+        # note: This warning is emitted in situations where the calling process has more
```
dpipemazo, line 589: :+1:

languages/python/atom/element.py
```diff
@@ -605,6 +615,7 @@ def command_loop(self, n_procs=1, block=True, read_block_ms=1000):
                 p.join()
 
     def _increment_command_group_counter(self, _pipe):
+        """Incremeents reference counter for element stream collection"""
```
dpipemazo, line 618: typo: `incremeents`

languages/python/atom/element.py
```diff
@@ -632,7 +644,7 @@ def _decrement_command_group_counter(self, _pipe):
             self._clean_up_streams()
         return result
 
-    def _command_loop(self, shutdown_event, read_block_ms=1000):
+    def _command_loop(self, shutdown_event, read_block_ms=10000):
```
dpipemazo, line 647: :+1:

languages/python/atom/element.py
```diff
@@ -697,8 +709,13 @@ def _command_loop(self, shutdown_event, read_block_ms=1000):
             if not cmd_responses:
                 continue
             cmd_stream_name, msgs = cmd_responses[0]
-            assert cmd_stream_name.decode() == stream_name, \
-                "Expected received stream name to match: %s %s" % (cmd_stream_name, stream_name)
+            if cmd_stream_name.decode() != stream_name:
```
dpipemazo, line 712: :+1:

languages/python/atom/element.py
```diff
@@ -781,8 +798,15 @@ def _command_loop(self, shutdown_event, read_block_ms=1000):
             kv["cmd_id"] = cmd_id
             kv["element"] = self.name
             kv["cmd"] = cmd_name
-            _pipe.xadd(self._make_response_id(caller), kv, maxlen=STREAM_LEN)
-            _pipe.execute()
+            try:
+                _pipe.xadd(self._make_response_id(caller), kv, maxlen=STREAM_LEN)
+                _pipe.execute()
+            except:
+                # If we fail to xadd the response, go ahead and continue
```
dpipemazo, line 805: :+1:

dpipemazo:
Looks good, let's merge when you feel it's ready.

Issue #342, opened 2020-06-25
Update `entry_read_n` API to take optional `last_id` argument


dpipemazo:
https://github.com/elementary-robotics/atom/blob/master/languages/python/atom/element.py#L1033

I believe can replace the `-` in the `xrevrange` call with the last ID we've seen. In this way we'll get blank data if there's no new data since we last read. This will help us with error-checking when we get the most recent piece of data from a stream.

PR #343, opened 2020-06-26
add last_id argument to entry_read_n

mvanveen:
Implements #342

### DIFF
--- a/languages/python/atom/element.py
+++ b/languages/python/atom/element.py
@@ -1,4 +1,5 @@
 import copy
+from collections import defaultdict
 from multiprocessing import Process
 import multiprocessing
 from traceback import format_exc
@@ -104,6 +105,9 @@ class Element:
             # Keep track of command_last_id to know last time the element's command stream was read from
             self.command_last_id = _pipe.execute()[-1].decode()
             _pipe = self._release_pipeline(_pipe)
+
+            # Keep track of stream last_id \forall streams
+            self._entry_read_n_last_id = defaultdict(int)
 
             # Init a default healthcheck, overridable
             # By default, if no healthcheck is set, we assume everything is ok and return error code 0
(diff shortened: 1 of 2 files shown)


languages/python/atom/element.py
```diff
@@ -1042,22 +1047,32 @@ def entry_read_n(self, element_name, stream_name, n, serialization=None, force_s
...
+        if last_id:
+            kwargs.update({'min': last_id})
+        uid_entries = self._rclient.xrevrange(stream_id, **kwargs)
+
         for uid, entry in uid_entries:
             entry = self._decode_entry(entry)
             serialization = self._get_serialization_method(entry, serialization, force_serialization, deserialize)
             entry = self._deserialize_entry(entry, method=serialization)
             entry["id"] = uid.decode()
+            self._entry_read_n_last_id[stream_id] = int(entry['id'].split('-')[0])
```
dpipemazo, line 1075: I think the entries come back in reverse -- you may wind up with the oldest of the entries if you read back a chunk here? Has this been tested?

languages/python/atom/element.py
```diff
@@ -105,6 +106,9 @@ def __init__(self, name, host=None, port=DEFAULT_REDIS_PORT,
             self.command_last_id = _pipe.execute()[-1].decode()
             _pipe = self._release_pipeline(_pipe)
 
+            # Keep track of stream last_id \forall streams
+            self._entry_read_n_last_id = defaultdict(int)
```
dpipemazo, line 110: This wasn't quite in the spec and I'm not sure this is the behavior we want -- I prefer giving users the explicit ability to handle their destiny. A user very well might want to read the same thing twice and this is some magic under the hood that would prevent that. If the user wants to track the last ID they can do it themselves and pass it in. Please remove.

languages/python/tests/test_atom.py
```diff
@@ -1210,6 +1210,55 @@ def test_timeout_ms(self):
 
         assert int(round(diff, 2)) == 2
 
+    def test_test_entry_read_n_last_id(self, caller, responder):
```
dpipemazo, line 1213: :+1: tests look good

languages/python/atom/element.py
mvanveen: oof thanks for calling that out re: ordering.  It has been tested to ensure that the sequence id is updating and increasing, but not in a way that would have surfaced this issue within one `entry_read_n` call.

languages/python/atom/element.py
mvanveen: Sure thing- I wasn't sure if the intent was to keep a running `last_id` or have the user supply one so wanted to initially accommodate both quickly.  Agree that this could be magic that is not wanted and will remove :+1:

Issue #353, opened 2020-07-27
Add Metrics Support


dpipemazo:
Generally want to be able to add metrics to the Atom ecosystem and get decent performance. We're looking for high-fidelity data when connected to the metrics instance within atom as well as the ability to send the data to a longer-term data store that can be queried to fill cloud dashboards.

- Add a metrics server to the nucleus of some sort 
- Add supervisord to launch both redis + metrics within the nucleus
- Choose metrics server based on possible integrations -- needs to be able to integrate with off-the-shelf enterprise cloud monitoring services as a source
- Add/design APIs within atom to add metrics

Generally excited to start discussing this.

dpipemazo:
[Prometheus](https://prometheus.io/) seems to be favorite project in the space these days after a surface-level dive. This is in comparison to [statsd](https://github.com/statsd/statsd) which was previously the recommended tool.

It seems that Prometheus is different from statsd in a few ways:
1. Pull-type metrics architecture instead of push. Each client logging metrics logs them to their internal memory which then gets polled by the server at some interval. Statsd, on the other hand, has clients send a UDP message to the statsd server with each item being logged. In theory Prometheus should be more performant in critical paths though perhaps at higher complexity due to the pull-based model.
2. Built-in time-series database. Prometheus integrates a [local, on-disk time-series DB](https://prometheus.io/docs/prometheus/latest/storage/) with the option to integrate with remote storage through integrations. We could mount a folder into the docker container (like we do with postgres in some places) and then get persistent time-series DB on the robots that lasts across stack launches. Statsd is only a metrics server and would lose all of its data when the stack when up/down to my knowledge. Statsd is typically used with something like Graphite in order to get time-series data storage.

Definitely more work to do on the investigation -- just wanted to leave this here and note I currently lean a bit towards prometheus.

dpipemazo:
[Redis time series](https://oss.redislabs.com/redistimeseries/) is another appealing option -- get to continue to focus on the Redis we already know and it has a grafana plugin. 

After looking at Prometheus and its pull-based model a bit more I'm not entirely sure it's right for Atom, especially given the [trouble with multithreading](https://github.com/prometheus/client_python) they mention in their Python client docs. 

Redis time-series is pushed based, allows you to specify per-key aggregation/downsampling rules and likely will work for us out of the box. Going to play around with it a bit to see how it goes.

Issue #351, opened 2020-07-23
Add environment variables to control redis connection


dpipemazo:
We want to agree on and implement in the Atom clients a set of environment variables which dictate which redis we connect to as our "nucleus" for a given element as well as which method we use (TCP vs socket). These environment variables would only be used in the event that explicit arguments aren't passed to the element constructor as these are already arguments to the constructor.

If we implement this in the element spec we can then easily point our docker containers at different hosts through standardized docker-compose variables which would be quite nice. 

@Cody-G

Cody-G:
Sounds great!  As we discussed, one immediate benefit of this is that we can point `stream-viewer` at an image stream on a remote host, though there are plenty of other use cases.

PR #359, opened 2020-08-16
353/metrics redis time series

dpipemazo:
Working on #353 

As of now, this PR:

1. Adds in `supervisord` for the nucleus
2. Launches 2x copies of redis in the nucleus using `supervisord` -- the normal "Atom" copy running on port `6379` and/or `redis.sock` and then a new "metrics" instance running RedisTimeSeries + persistent backups on port `6380` and/or `metrics.sock`. 

Rationale behind using 2x redises for this is as such:
- We want to be wary of extra data bogging down the single-threaded redis in the critical path
- We want backup/snapshots of the time-series database but don't want to add this overhead to the critical path. This is a redis-wide setting
- We want to load the time-series module and don't want to take risk on its performance impacts in the critical path
- There's really not much additional overhead here -- it's essentially just "sharding" the metrics data and we should have the CPU overhead to handle it.

Confirmed everything launching OK, started playing around with redis time series for metrics: https://oss.redislabs.com/redistimeseries/commands/

It's a fairly simple but pretty powerful interface that should work quite well for us. Still need to work on the spec to wrap the time-series APIs in atom APIs and also add/install/use the [redis time series Python client](https://github.com/RedisTimeSeries/redistimeseries-py)


dpipemazo:
Functionality complete with decent unit tests to get the metrics data into a redis time series instance. Still needs the following:

1. End -> End integration testing using live elements in our system
2. Documentation on the metrics spec
3. Documentation for metrics APIs in the python client
4. Documentation on grafana and perhaps some default dashboards
5. A decent system-level grafana source that can help us track CPU and other things?

Can be tested using grafana in a docker-compose file:
```
  grafana:
    container_name: grafana
    image: grafana/grafana
    environment:
      - "GF_INSTALL_PLUGINS=redis-datasource"
    ports:
      - "3001:3000"
```

Pull up `localhost:3001` and then use the default credentials (admin/admin) to log in. You should be able to add the metrics database as a data source using `redis://<container/service name>:6380`. You can then make a dashboard using the `RANGE` redis time series query and see your metrics come to life! Pretty slick implementation by the redis labs team on both the feature and the grafana plugin!

dpipemazo:
Another cool thing to note with this feature is that the metrics DB is persisting itself to disk roughly every 5 minutes or so (or after a significant amount of writes). It would be a quick/easy win to add a cron job in the nucleus container or some system script to upload this backup to some data sink (s3, gcp, azure, etc.) regularly s.t. we (optionally) have snapshots of the state of the metrics backed up in case the system goes offline and is unreachable for whatever reason. We can then boot up a redis using this backup file and have the exact state of the system!

Also worth noting is that the above file is being written to `/metrics/` in the nucleus docker container. If you mount some folder on your system into the container here you'll get time-series data that persists across stack reboots (down to the few minutes or so that may be missing between syncs). This is pretty nifty and will let us easily keep quite a decent amount of history on the edge with atom.

dpipemazo:
After a bit more work this is ready for review. Want to hold off on merging until we run some end to end tests off of the dev build, but we now have the following behavior:

1. Metrics are default off. Set `ATOM_USE_METRICS` to `TRUE` at the time that an element is initialized in order to attempt to use metrics. This will enable us to not break old setups.
2. Even if metrics is turned on, if we detect in init that we can't connect to a metrics server we, by default, simply disable metrics and make all metrics calls no-ops. The metrics calls will return non-truthy values in this case but won't add any undue latency. This will enable us to instrument elements without worrying about breaking/porting older nucleuses without the metrics server.
3. If you want metrics functionality to be enforced/enabled, you can pass a flag, `enforce_metrics` to the element creator. This will enforce that the metrics server is present and throw an error if not. Eventually this should become default true but is left default false for now, again, to avoid breaking existing setups.

It is expected for elements to call `metric_create` to set up their labels and rules in their setup. However, with time-series data being persistent the metric will only need to truly be created the first time. As such, we now have an `update` flag on `metric_create` where it won't error out if the metric already exists but will instead overwrite the retention, labels and rules without ever deleting the metric or having any weird behavior while in transition like the metric not existing in the DB. This allows reads to be guaranteed to succeed across any call to `metric_create` which is important for stability. It's generally expected to set this flag to True, but it is explicitly not the default as people should be aware of what they're doing and calls shouldn't update silently IMO.

Unit tests are updated across the board and are pretty thorough at this point.

dpipemazo:
A few changes to the API and now I think it's ready for some integration testing with existing elements:
- `metric_add` now takes a variadic list of tuples. It uses Redis Time series's bulk addition endpoint so it can do this in a single transaction. First value in the tuple is a key, second is the value. Third, optionally, is the timestamp. If timestamp isn't specified we'll assume `'*'`, unless `use_curr_time` is true or `execute` is false in which case we'll grab the wallclock timestamp.
- `metric_add` uses `madd` instead of `add`. This is safer as it forces us to explicitly use `metric_create` and won't implicitly create on `add` if the metric doesn't exist. Redis has a flag coming out for `add` to disable this behavior in the future but this path will force us into the paradigm of creating metrics at element init and then just adding to them through runtime.

dpipemazo:
Added metrics logging internally to atom on:
- command loop (per worker)
- command send (per element/command pair)
- command handlers (per handler)
- entry read since
- entry read n
- entry write
- reference create
- reference create from stream
- reference get

We're tracking errors, timing (serialization, deserialization, redis blocking, command length) and counts across most of what could possibly be found interesting -- approximately 25 metrics in all. Each metric will have the following data frequency and retention:
- All data for 24H
- Data bucketed by 10m for 3 days
- Data bucketed by 1h for 30 days
- Data bucketed by 1d for 365 days
All downsampling is done off of the original stream to not lose fidelity.

Full docs/details will come later, for now excited to rebuild a few elements in our stack on top of this and see what we can see on the dashboards.

Dockerfile
```diff
@@ -114,12 +125,16 @@ FROM atom as nucleus
 
 # Add in redis-server
 COPY --from=atom-source /usr/local/bin/redis-server /usr/local/bin/redis-server
+# Add in redis-time-series
+COPY --from=atom-source /atom/third-party/RedisTimeSeries/bin/redistimeseries.so /etc/redis/redistimeseries.so
```
mvanveen, line 129: random q: why `/etc` instead of, say, `/lib`?

config/nucleus/redis/redis-metrics.conf
```diff
@@ -0,0 +1,1813 @@
...
+#
+#   It is also possible to remove all the previously configured save
+#   points by adding a save directive with a single empty string argument
+#   like in the following example:
+#
+#   save ""
+
+save 900 1
+save 300 10
+save 60 10000
```
mvanveen, line 289: should we tune this in practice?

docker-compose-dev.yml
```diff
@@ -16,6 +16,19 @@ services:
         target: /shared
         volume:
           nocopy: true
+      - "./:/metrics"
+    ports:
+      - "6379:6379"
+      - "6380:6380"
+      - "6080:6080"
+
+  grafana:
```
mvanveen, line 25: :balloon: this opens the door to all sorts of fun observability shenanigans possibly down the road

languages/python/requirements.txt
```diff
@@ -1,2 +1,3 @@
 lazycontract==0.9.6
 msgpack==0.6.2
+redistimeseries
```
mvanveen, line 3: is there a particular version we're using that's worth freezing

mvanveen:
I skimmed- must admit I have not managed to track down the metrics configuration mentioned in your recent comments in this PR yet.  I did make out a few devopsy kinds of comments my read- please feel free to ignore.

I think sharding seems like a great strategy here to compartmentalize concerns.  I'm excited to see grafana in this PR and interested to see how that pertains to both this use case and future monitoring use cases.

The backup strategy you mentioned in an earlier sounds great- do we have a ticket for it yet?

excited to see this in production!

dpipemazo:
After testing a bit I saw some weird behavior and then found that [redis-py pipelines are not safe to be shared between threads](https://github.com/andymccurdy/redis-py#thread-safety). We must be getting away with what we're doing with the pipeline pool due to the fact we're resetting it between each use. The current async flush design assumes that they are thread-safe and it's not too bad of a change to move to a paradigm where for async metrics you simply get your own pipeline and pass it to all of your `metrics_add` calls and flush it yourself. This is actually more deterministic time-wise (as you may never be stuck with timing delays due to flushing someone else's metric payload due to timing conditions, and is likely a better solve anyway.

dpipemazo:
> The backup strategy you mentioned in an earlier sounds great- do we have a ticket for it yet?

Not yet -- that'll go on one of our configuration-as-code repos. In atom here the backups will be 
persisted to disk automatically in this ticket due to the redis saving settings you found.

languages/python/requirements.txt
dpipemazo: good call-out -- I usually finish testing and then `pip freeze` to lock it in.

docker-compose-dev.yml
dpipemazo: yeah I'm pretty excited about this. The folks at redis labs made a super nice redis plugin for grafana that's now shipped with it that works with the redis time series DB

config/nucleus/redis/redis-metrics.conf
dpipemazo: yes, absolutely. We could even expose these as env variables and then use `sed` or something in the launch script to write them into the conf file based on user preference.

dpipemazo:
- Fixed unit tests
- Added in metrics documentation
- Added in auto-generation of grafana dashboards at boot. This performs service discovery for all elements running the new metrics system and makes all of the dashboards for the elements found.
- Added in redis monitoring dashboards for both the nucleus and metrics. These are just the off-the-shelf redis datasource dashboard but still much better than what we had previously
- Added in a default dashobard that shows you lists you can click on for all of the above dashboards
- Added build of new `metrics` container to CircleCI job. Passing/working for intel, not sure yet if ARM will work and haven't put much time into it. Would be OK just not running the ARM build of this for now if it continues to fail and just come back to it.

In order to make the dashboards above we now have a small python script `create_dashboards.py` which uses `jinja2` and the grafana HTTP API to parse some template configs and update grafana to match the configs at boot.

config/nucleus/redis/redis-metrics.conf
```diff
@@ -0,0 +1,1813 @@
...
+# If a pid file is specified, Redis writes it where specified at startup
+# and removes it at exit.
+#
+# When the server runs non daemonized, no pid file is created if none is
+# specified in the configuration. When the server is daemonized, the pid file
+# is used even if not specified, defaulting to "/var/run/redis.pid".
+#
+# Creating a pid file is best effort: if Redis is not able to create it
+# nothing bad happens, the server will start and run normally.
+pidfile /var/run/redis_6379.pid
```
nceruchalu, line 227: This is using the same pidfile as redis-atom.conf
Wouldn't this be a problem?

dpipemazo: ah, yes. Good catch!

dpipemazo:
Added `monitor`, a metrics element that uses the `psutil` to scrape system metrics and log them into redis time series similarly to the other metrics.

We are now logging (in configurable intervals, typically around ~10s):
- CPU (per-CPU, per-process and per-thread)
- Disk space
- Memory 
- Network
- File descriptors
- Temperature sensors (if present in chipset)

Working on adding in a bunch more default grafana dashboards to visualize all of the data

dpipemazo:
Now at a feature-complete state (minus docs). We now have:

1. A home dashboard which shows you most of what you need to see. It has links to all of the other dashboards
2. Detailed dashboards for cpu, disk, network and memory

Grafana server has been moved to port `3001` by default so as not to conflict with other things that typically run on `3000`.

Take it for a spin and see how it monitors your system!

```
git submodule update --init --recursive
docker-compose -f docker-compose-dev.yml build
docker-compose -f docker-compose-dev.yml up -d
```

Then, after giving it 10s or so to boot, take a look at `localhost:3001` and log in with `admin` username and `admin` password. You can then see how the system is monitoring your computer

One thing I still want to add in here is monitoring for certain directories to keep an eye on growing/large files. I think that's the real main thing it is missing to give us good peace-of-mind in the field.

dpipemazo:
Now that we're launching the nucleus with `supervisor` we need the ability in the elements to wait until the nucleus is ready before launching the elements themselves. Previously we were rslying on the `requires` field in the docker-compose and were getting away with it. When building this out it became apparent we needed to know at the container-level where we were going to connect to the nucleus so I went ahead and addressed https://github.com/elementary-robotics/atom/issues/351 .

We now have the following variables and defaults in the Dockerfile:
```
ENV ATOM_NUCLEUS_HOST ""
ENV ATOM_METRICS_HOST ""
ENV ATOM_NUCLEUS_PORT "6379"
ENV ATOM_METRICS_PORT "6380"
ENV ATOM_NUCLEUS_SOCKET "/shared/redis.sock"
ENV ATOM_METRICS_SOCKET "/shared/metrics.sock"
```

The Python element has been updated to use these by default but otherwise can still be configured with any type of connection desired to override. No code changes needed, all default behavior is the same.

We also now have a script in `/usr/local/bin/wait_for_nucleus.sh` which will read these vars and send a `PING` to the redis server, waiting for a `PONG` response. The script then calls `exec` on any passed command-line arguments in a similar fashion to https://github.com/eficode/wait-for

Our default launch command should now be `/usr/local/bin/wait_for_nucleus.sh python3 <my_element.py` or similar.

dpipemazo:
- Updated `RedisTimeSeries` to the latest version to pull in bugfixes that were impacting grafana
- Now installing `redistimeseries-py` from source in order to be compatible with the newly updated version of `RedisTimeSeries`, had blocking bug otherwise. 
- Updated Atom code/tests to account for updating `redistimeseries-py`. Calling of non-timeseries redis commands has changed. Nothing too major.
- Updated Atom tests to account for new RedisTimeSeries behavior on writes of the same timestamp. New behavior is overwrite, not error. This is somewhat better than the previous behavior but still can lead to data loss. 
- Updated grafana redis datasource to pull in new capability to connect to redis over Unix socket
- Updated `docker-compose-dev.yml` for bugfix w.r.t. running metrics for the first time and `metrics.dbm` not existing.

dpipemazo:
- Improved/tested `metrics_add_type` API
- Improved `metrics_monitor.py` to use `metrics_add_type`
- Updated formatting to pass `.flake8`
- Added ability + documentation for users to create, save, and load their own dashboards with no code changes. See `metrics/dashboards/user/README.md`

languages/python/atom/element.py
```diff
@@ -1440,3 +2103,448 @@ def reference_get_timeout_ms(self, key):
...
+
+    def metrics_create(
+        self,
+        level,
+        m_type,
+        *m_subtypes,
+        retention=METRICS_DEFAULT_RETENTION,
+        labels={},
+        agg_timing=METRICS_DEFAULT_AGG_TIMING,
+        agg_types=[]):
```
mvanveen, line 2326: might want to make these default to `None` instead of `[]` and handle the construction of defaults for `agg_types`, `labels` etc. into the actual method.

The reason behind this is these variables will be constructed once at module time and can be mutated in ways that affect subsequent invocations worst case.

dpipemazo: Nice catch and good call!

languages/python/atom/element.py
```diff
@@ -1440,3 +2103,441 @@ def reference_get_timeout_ms(self, key):
...
+        _rules = {}
+        for agg in agg_types:
+            for timing in agg_timing:
+                _rule_key = self._make_metric_id(self.name, m_type, *m_subtypes, agg)
+                _rules[f"{_rule_key}:{timing[0]//(1000 * 60)}m"] = (agg, timing[0], timing[1])
+
+        return self.metrics_create_custom(level, _key, retention=retention, labels=_labels, rules=_rules)
+
+    def metrics_add(self, key, val, timestamp=None, pipeline=None, enforce_exists=True, retention=86400000, labels={}):
+        """
```
MeganParadowski1, line 2387: A future improvement could be to add a method that takes in multiple samples and series, utilizing madd with more than one key/timestamp/value pair

dpipemazo: Yes -- I actually had this working in the original implementation but the API wound up being a bit clunky so I rolled it back. We can get similar performance for now through using a pipeline with multiple ADD calls -- will yield a single redis transaction but will make multiple redis command calls. Not sure how much MADD is better than pipelined ADDs.

Cody-G:
While running our tests of the Basler element with this version of Atom I'm getting socket timeouts like this:
```
self = <redis.connection.HiredisParser object at 0x7f607b78b898>, timeout = <object object at 0x7f6086e6a880>, raise_on_timeout = True

    def read_from_socket(self, timeout=SENTINEL, raise_on_timeout=True):
        sock = self._sock
        custom_timeout = timeout is not SENTINEL
        try:
            if custom_timeout:
                sock.settimeout(timeout)
            if HIREDIS_USE_BYTE_BUFFER:
>               bufflen = recv_into(self._sock, self._buffer)

/opt/venv/lib/python3.7/site-packages/redis-3.4.1-py3.7.egg/redis/connection.py:421: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

sock = <socket.socket [closed] fd=-1, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0>
args = (bytearray(b'*1\r\n*2\r\n$13\r\nresponse:test\r\n*1\r\n*2\r\n$15\r\n1598983779396-0\r\n*6\r\n$7\r\nelement\r\n$6\r\nba...x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'),)
kwargs = {}

    def recv_into(sock, *args, **kwargs):
>       return sock.recv_into(*args, **kwargs)
E       socket.timeout: timed out

/opt/venv/lib/python3.7/site-packages/redis-3.4.1-py3.7.egg/redis/_compat.py:75: timeout
```

@dpipemazo I think you added more control of socket timeouts recently to fix an issue, probably I just need to change the default timeout?

dpipemazo:
@Cody-G we did add timeouts but we shouldn't be running into them. This means that the transfer of data/timeouts is occasionally getting into the level of seconds here which doesn't seem quite right. Perhaps that does make sense.

You can set `data_timeout` when creating the element to something > the default 5s, or I believe to None to turn it off: https://github.com/elementary-robotics/atom/blob/latest/languages/python/atom/element.py#L37

Cody-G:
@dpipemazo thanks!  After looking into this a bit further I now think this problem is not due to the metrics PR, but due to multiprocessing in `command_loop`.  I think the quick fix will be for me to call `_command_loop` as you did with spec-robot recently.

languages/python/atom/config.py
```diff
@@ -22,3 +23,48 @@
...
+    # Keep data in 10m buckets for 3 days
+    (600000,  259200000),
+    # Then keep data in 1h buckets for 30 days
+    (3600000, 2592000000),
+    # Then keep data in 1d buckets for 365 days
+    (86400000, 31536000000),
+]
+
+# Metrics logging levels
+class MetricsLevel(Enum):
```
Cody-G, line 61: Minor thing, could we match theses enums with the enums we're using for `LogLevel`?
See here:
https://github.com/elementary-robotics/atom/blob/9a455e247728e0ffe831a82069c2b21d7b2ee2de/languages/python/atom/messages.py#L167-L174

The only discrepancies I see are `ERR` vs `ERROR` and `WARN` vs `WARNING`

dpipemazo: We definitely could and it would make things a bit easier. There's one additional level on metrics: `TIMING` which I thought might be useful. Happy to make them match if there's more value there.

dpipemazo: Ah -- you're mainly talking about using full strings instead of partials? Sure. I always mess this up. We should ideally match the [syslog conventions](https://linux.die.net/man/3/syslog) but I see this will result in changes in multiple places.

Cody-G: Exactly, just the strings.  I don't mind whether we use full vs partial, just prefer that the `LogLevel` strings match the `MetricsLevel` strings because it'll be hard to remember.

Cody-G: (and I do like having `TIMING`)

PR #364, opened 2020-09-07
362/standardize logging

MeganParadowski1:
This PR implements the following logic to standardize logging in order to better integrate with any log management service provider: 
- Mounts a shared logging volume to all docker containers at /var/log/atom
- Implements a RotatingFileHandler logger to silently and automatically rotate the log file when it gets to a given size (currently 2000 bytes)
- Removes the stdout log function and outputs logs with the element logger

Future improvements: 
- Convert the log files to json so that we can log extra parameters as key value pairs as opposed to strings. This will make them much easier to query for in a log management service.


.circleci/docker-compose-circle.yml
```diff
@@ -11,6 +11,7 @@ services:
         target: /shared
         volume:
           nocopy: true
+      - log:/var/log/atom
```
dpipemazo, line 14: Interesting syntax here as this isn't how we do it for the `/shared` -- this definitely uses the `tmpfs` and not a local folder named `log`?

languages/python/atom/element.py
```diff
@@ -126,12 +134,22 @@ def __init__(self, name, host=ATOM_NUCLEUS_HOST, port=ATOM_NUCLEUS_PORT, metrics
...
+        #
+        # Set up logger
+        # 
+        rfh = logging.handlers.RotatingFileHandler(f'{ATOM_LOG_DIR}{self.name}.log', maxBytes=2000)
+        extra = {'element_name': self.name}
+        formatter = logging.Formatter("%(asctime)s %(element_name)s [%(levelname)s]: %(message)s")
+        rfh.setFormatter(formatter)
+        logger.addHandler(rfh)
+        self.logger = logging.LoggerAdapter(logger, extra)
+        self.logger.setLevel(os.getenv("ATOM_LOG_LEVEL", "INFO"))
```
dpipemazo, line 151: This is making a bit of an assumption that the user set a valid level. We should error-check on that and correct for them, i.e. if they set something invalid we fall back to "INFO". A fallback to a default level would be a better behavior than a crash/error due to the user setting something invalid.

languages/python/atom/element.py
```diff
@@ -1732,38 +1730,6 @@ def entry_write(self, stream_name, field_data_map, maxlen=STREAM_LEN, serializat
 
         return ret[0].decode()
 
-    def log(self, level, msg, stdout=True, _pipe=None, redis=False):
```
dpipemazo, line 1735: Please keep this for legacy purposes. When users use this API you should translate the calls into the right `self.logger` calls. This will give us free upgrades and we can encourage people to stop using it in a future release.

dpipemazo (changes requested):
Overall looks good. Please handle invalid log levels and please don't deprecate `element.log` but instead forward all calls to `element.log` through the new logging utility.

.circleci/docker-compose-circle.yml
MeganParadowski1: Yes, tested on the container with 
`# stat --file-system --format=%T /var/log/atom
tmpfs`

languages/python/atom/element.py
MeganParadowski1: Good point - will update!

dpipemazo:
@MeganParadowski1 Please also add parameterization for the log file size -- ideally through an env variable s.t. different elements can change their rotation size if they'd like.

languages/python/atom/element.py
```diff
@@ -126,12 +134,22 @@ def __init__(self, name, host=ATOM_NUCLEUS_HOST, port=ATOM_NUCLEUS_PORT, metrics
...
+        # Set up metrics logging levels
         #
-        self._log_level = LogLevel[os.getenv("ATOM_LOG_LEVEL", "INFO")]
         self._metrics_level = MetricsLevel[os.getenv("ATOM_METRICS_LEVEL", "TIMING")]
         self._metrics_use_aggregation = os.getenv("ATOM_METRICS_USE_AGGREGATION", "FALSE") == "TRUE"
 
+        #
+        # Set up logger
+        # 
+        rfh = logging.handlers.RotatingFileHandler(f'{ATOM_LOG_DIR}{self.name}.log', maxBytes=2000)
```
dpipemazo, line 145: Please make this `2000` a parameter, set by ENV

dpipemazo: Please also perform some escaping (if necessary) on the element name to ensure it complies with the character rules for filenames. I believe the user could currently set an element name that would break this.

languages/python/atom/element.py
```diff
@@ -42,8 +42,9 @@
 ATOM_NUCLEUS_SOCKET = os.getenv("ATOM_NUCLEUS_SOCKET", DEFAULT_REDIS_SOCKET)
 ATOM_METRICS_SOCKET = os.getenv("ATOM_METRICS_SOCKET", DEFAULT_METRICS_SOCKET)
 
-# Get the log directory
+# Get the log directory and log file size
 ATOM_LOG_DIR = os.getenv("ATOM_LOG_DIR", "")
+ATOM_LOG_FILE_SIZE = os.getenv("ATOM_LOG_FILE_SIZE", DEFAULT_LOG_FILE_SIZE)
```
dpipemazo, line 47: I think you're going to need an `int` around this?

languages/python/atom/element.py
```diff
@@ -1730,6 +1744,24 @@ def entry_write(self, stream_name, field_data_map, maxlen=STREAM_LEN, serializat
...
+            stdout (bool, optional): Whether to write to stdout or only write to log stream.
+            _pipe (pipeline, optional): Pipeline to use for the log message to
+                be sent to redis
+            redis (bool, optional): Default true, whether to log to
+                redis or not
+        """
+
+        numeric_level = getattr(logging, level.upper(), None)
+        if not isinstance(numeric_level, int):
+            numeric_level = 20
```
dpipemazo, line 1762: What does `20` mean here? Please avoid "magic" numbers where possible, either through thorough commenting or constants.

MeganParadowski1: It's the numeric log level for INFO - I'll update

languages/python/atom/element.py
MeganParadowski1: You're right, I set it in the dockerfile as an int, but if the user sets it then it may be a string

languages/python/atom/config.py
```diff
@@ -66,3 +66,7 @@ class MetricsLevel(Enum):
     INFO = 6
     TIMING = 7
     DEBUG = 8
+
+# Logging
+DEFAULT_LOG_FILE_SIZE = 2000
```
dpipemazo, line 71: Slight semantic note -- I typically prefer to have constants start with their "module"/functionality to make it a bit easier to surface what it pertains to. Here, we could do `LOG_DEFAULT_FILE_SIZE` and `LOG_DEFAULT_LEVEL` to help group these under the logical `LOG` "module"/functionality. This applies throughout to functions, etc.

languages/python/atom/element.py
```diff
@@ -126,12 +135,33 @@ def __init__(self, name, host=ATOM_NUCLEUS_HOST, port=ATOM_NUCLEUS_PORT, metrics
...
+        # Set up logger
+        # 
+        try: 
+            rfh = logging.handlers.RotatingFileHandler(f'{ATOM_LOG_DIR}{self.name}.log', maxBytes=ATOM_LOG_FILE_SIZE)
+            extra = {'element_name': self.name}
+            formatter = logging.Formatter("%(asctime)s element:%(element_name)s [%(levelname)s] %(message)s")
+            rfh.setFormatter(formatter)
+            logger.addHandler(rfh)
+            self.logger = logging.LoggerAdapter(logger, extra)
+        except Exception as e:
```
dpipemazo, line 153: Generic try/except is generally not a great strategy. We use it in certain places but I believe the better way to handle this is to catch the errors we know about/can reasonably happen explicitly and let it throw in the event we hit a new one.

One downside of this way of doing it is that you lose the full stack trace as well which is quite useful. 

Finally, putting this much code in a single `try/except` is also not ideal -- they should be small and specific to known errors in each case.

languages/python/atom/element.py
```diff
@@ -1734,8 +1744,7 @@ def entry_write(self, stream_name, field_data_map, maxlen=STREAM_LEN, serializat
 
     def log(self, level, msg, stdout=True, _pipe=None, redis=False):
         """
-        Writes a message to log stream with loglevel.
-
+        Forwards calls to self.logger
```
dpipemazo, line 1747: :+1:

dpipemazo (changes requested):
Looking good.

Please tighten up the error-handling in the logging initialization but otherwise in great shape!

PR #365, opened 2020-09-16
Cpp client rewrite - work in progress

BarbBoyaji:
not final!


aywaldron:
Just leaving a comment here to track some things Barbara and I discussed today. 

Currently, the redis command wrappers are returning a generic `redis_reply` type from which the parsed reply data must then be extracted. From my understanding the `redis_reply` type serves as a generalized way to manage the memory of the buffers returned by the redis commands. We talked about making the reply data types into their own classes (there are 3 types currently), each with their own cleanup method, and having their data returned directly by the redis command wrappers to simplify the code and logic where the wrappers are called and the memory cleanup method determined. Then, for example, an `xrevrange` call would return an `entry_response` object which contained the response data in the expected format and on which `cleanup` could be called to clear out the buffer. Then the generic `redis_reply` type may not be necessary or may be a parent class to the individual response types.

PR #370, opened 2020-10-01
361/formatting

dpipemazo:
**Infrastructure Changes**

- Creates new `formatter` container which contains `flake8` and `black`
- Adds build + deploy of `formatter` to CircleCI
- Adds formatting check to **all** python code in the Atom repo
- Adds documentation on how to use/check formatting to the README

**Code Changes**
- Formats code with `black`
- Updates code to comply with ER `.flake8`

### DIFF
--- a/languages/python/atom/__init__.py
+++ b/languages/python/atom/__init__.py
@@ -1,4 +1,4 @@
-from atom.element import Element, AtomError, MetricsLevel
+from atom.element import Element, AtomError, MetricsLevel  # noqa F401
 from atom.config import VERSION
 
 __version__ = VERSION
--- a/.circleci/Dockerfile-flake8
+++ b/.circleci/Dockerfile-flake8
@@ -1,3 +1,0 @@
-FROM alpine/atom-flake8:3.7.0
-
-RUN apk add git openssh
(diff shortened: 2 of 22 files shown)


plipay (approved):
Looks great, thanks for taking the initiative on this.
I'd maybe also toss in the docker run cmd example you mentioned on slack earlier into the README: `docker run -e DO_FORMAT=y --mount src=$(pwd),target=/code,type=bind elementaryrobotics/formatter` , so there's an easy copy-paste 1-liner folks can do without having to clone the repo

languages/python/atom/__init__.py
```diff
@@ -1,4 +1,4 @@
-from atom.element import Element, AtomError, MetricsLevel
+from atom.element import Element, AtomError, MetricsLevel  # noqa F401
```
Cody-G, line 1: Is this comment something that the formatter added?

Cody-G (approved):
Looks great, agreed with Peter would be nice to have that docker run command in the README as well.  I just marked one weird comment, wondering if that was intentional or added by Black

languages/python/atom/__init__.py
dpipemazo: No, I added this since flake8 complains about it being unused otherwise

Cody-G: Ah cool, didn't know we could exclude with comments like that

dpipemazo: Yeah, also not something to make a habit of using :smile:

PR #374, opened 2020-10-13
Throw atom error on empty string in metrics labels

MeganParadowski1:


dpipemazo (approved):
:+1: looks great, thanks for also adding a test!

PR #376, opened 2020-10-21
Deduplicate log messages

MeganParadowski1:
When multiple instances of an element utilize the element logger, a new rotating file handler is added to the logger each time. This was causing duplicate log messages to be outputted to the ATOM_LOG_DIR.

### DIFF
--- a/languages/python/atom/element.py
+++ b/languages/python/atom/element.py
@@ -211,7 +211,7 @@ class Element:
             )
             rfh.setFormatter(formatter)
             logger.addHandler(rfh)
-            self.logger = logging.LoggerAdapter(logger, extra)
+            logger = logging.LoggerAdapter(logger, extra)
 
             #
             # Set up log level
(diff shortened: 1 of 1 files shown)


languages/python/atom/element.py
```diff
@@ -389,6 +368,37 @@ def __init__(
 
         self.logger.info("Element initialized.")
 
+    def _get_atom_logger(self, name):
+        # Avoid redefining the logger if it already exists
+        if loggers.get(name):
```
plipay, line 373: To be more explicit, I would do `if name in loggers:`

languages/python/atom/element.py
```diff
@@ -389,6 +368,37 @@ def __init__(
...
+            logger = logging.LoggerAdapter(logger, extra)
+
+            # Set log level
+            loglevel = os.getenv("ATOM_LOG_LEVEL", LOG_DEFAULT_LEVEL)
+            numeric_level = getattr(logging, loglevel.upper(), None)
+            if not isinstance(numeric_level, int):
+                loglevel = LOG_DEFAULT_LEVEL
+            logger.setLevel(loglevel)
+
+            loggers.update(dict(name=logger))
```
plipay, line 399: I would do `loggers[name] = logger`

plipay (changes requested):

plipay (approved):
Nice, much smaller diff now

dpipemazo (approved):

PR #380, opened 2020-10-23
element: don't use data timeout for unix socket connections

dpipemazo:
- Existing code depends on some longer-running blocking calls.
- This timeout was added to guard against faulty network
connections, we don't have this class of issues when connecting
via unix socket

### DIFF
--- a/languages/python/atom/element.py
+++ b/languages/python/atom/element.py
@@ -238,7 +238,6 @@ class Element:
                 self._metrics_socket_path = metrics_socket_path
                 self._mclient = RedisTimeSeries(
                     unix_socket_path=self._metrics_socket_path,
-                    socket_timeout=self._redis_data_timeout,
                     socket_connect_timeout=self._redis_connection_timeout,
                     client_name=self.name,
                 )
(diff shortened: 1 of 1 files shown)


MeganParadowski1 (approved):

Cody-G (approved):

plipay:
Just out of curiosity, what I recall having to do when we ran into timeout issues in vision processing, was explicitly setting         
```
element = Element(thread_id, data_timeout_ms=10000)
```
Not sure if this is relevant for unix sockets or not, but if there's a separate data timeout that is unnecessary on unix sockets, might make sense in this PR as well.

plipay (approved):
NVM my earlier comment, read into it more and what you have here makes sense.

dpipemazo:
@plipay yes, this PR removes setting the `data_timeout_ms` altogether for unix socket connections (most of what we do in our elements). It now only sets this option for connections over TCP.

Unix sockets will revert to the previous Atom behavior of not having a data timeout. The code as suggested above will ignore the option.

PR #384, opened 2020-11-11
Convert syslog loglevel

MeganParadowski1:
- Convert syslog log level to python log level to support the deprecated element.log() method
- Retain loggers by element name to avoid duplicating across multiple instances of an element

### DIFF
--- a/languages/python/atom/element.py
+++ b/languages/python/atom/element.py
@@ -63,7 +63,7 @@ from atom.config import (
 )
 from atom.config import MetricsLevel
 from atom.messages import Cmd, Response, StreamHandler, format_redis_py
-from atom.messages import Acknowledge, Entry, ENTRY_RESERVED_KEYS
+from atom.messages import Acknowledge, Entry, ENTRY_RESERVED_KEYS, LOG_LEVEL_CONVERSION
 import atom.serialization as ser
 
 # Need to figure out how we're connecting to the Nucleus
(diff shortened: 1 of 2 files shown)


languages/python/atom/element.py
```diff
@@ -79,6 +79,10 @@
 ATOM_LOG_DIR = os.getenv("ATOM_LOG_DIR", "")
 ATOM_LOG_FILE_SIZE = int(os.getenv("ATOM_LOG_FILE_SIZE", LOG_DEFAULT_FILE_SIZE))
 
+# Track loggers by element name to avoid duplication across multiple
+# instances of an element
+LOGGERS = {}
```
dpipemazo, line 84: This strategy won't hold across processes I think

languages/python/atom/messages.py
```diff
@@ -180,3 +181,13 @@ class LogLevel(Enum):
     NOTICE = 5
     INFO = 6
     DEBUG = 7
+
+
+# Convert syslog log levels to python log level
+LOG_LEVEL_CONVERSION = {
```
dpipemazo, line 187: :+1: this is clever.

dpipemazo (changes requested):
I am not sure the global-dict logger check will do us much good as these aren't shared in multi-process environments. 

Are we adding/do we have code to clean up the logger in the element deletion stack? That can also fix the problem and I'm not sure I see it in here.

MeganParadowski1:
> I am not sure the global-dict logger check will do us much good as these aren't shared in multi-process environments.
> 
> Are we adding/do we have code to clean up the logger in the element deletion stack? That can also fix the problem and I'm not sure I see it in here.

@dpipemazo I'm having a hard time determining where exactly the element is deleted. Could you point me in the right direction?

dpipemazo:
@MeganParadowski1 sure, in Python the __del__ method is called whenever the Python garbage collector/runtime deletes an element. 

https://github.com/elementary-robotics/atom/blob/latest/languages/python/atom/element.py#L429

This is where we put code that we want to run when an element is deleted.

PR #387, opened 2020-12-07
atom orb: add build labels

dpipemazo:
Adds in automatic build labeling to all Atom orb users (on orb upgrade). See README for info


aywaldron (approved):
Nice!

PR #390, opened 2020-12-10
Addition of new parameters API

### LINKED ISSUE #389: Allow for user specification of reference key
When writing a reference, the user should be able to specify the key to use if they want to. This will allow for general Redis key/val storage through atom.

aywaldron:
per the discussion in ticket #389, adds new parameters API for storing general key-val pairs in a Redis hash store. 

Also includes:
- Addition of user-specified reference keys
- Bug fix in `reference_delete` method
- Updated docker-compose files to work with most recent atom version
- Addition of submodule cloning to quickstart readme
- Some small documentation udpates/fixes

Resolves #389


languages/python/atom/element.py
```diff
@@ -590,15 +590,17 @@ def _metrics_validate_labels(self, labels):
...
         Args:
-
+            key (str, optional): User specified key; defaults to None
+        Returns:
+            Full reference key that includes element name
         """
-
-        return f"reference:{self.name}:{str(uuid.uuid4())}"
+        key = key if key else str(uuid.uuid4())
+        return f"reference:{self.name}:{key}"
```
dpipemazo, line 603: I think as designed right now if we want data passed between elements it will only work if the caller of the API has an element with the same name? This is fairly limiting. I think if we made it `reference:{key}` it would allow us to have any element be able to pull/write to a piece of data by name which is going to give us slightly better functionality.

languages/python/atom/element.py
```diff
@@ -2291,8 +2295,8 @@ def reference_create(
             # Serialize
             self.metrics_timing_start(self._reference_create_metrics["serialize"])
             for datum in data:
-                # Get the key name for the reference to use in redis
-                key = self._make_reference_id()
+                # Get the full key name for the reference to use in redis
+                key = self._make_reference_id(key)
 
                 # Now, we can go ahead and do the SET in redis for the key
```
dpipemazo, line 2301: Do we also need to modify the reference get API?

dpipemazo (changes requested):

languages/python/atom/element.py
aywaldron: `reference_get` takes in the entire reference key so it does not need to be modified if its able to know the full key used by `reference_create`.

languages/python/atom/element.py
```diff
@@ -2435,23 +2435,23 @@ def parameter_read(self, key, fields=None, serialization=None, force_serializati
 
             # Deserialize the data
             for field, val in data.items():
-                deserialized_data[field] = (
-                    ser.deserialize(val, method=serialization)
-                    if val is not None
-                    else None
-                )
+                if field != b"override":
```
dpipemazo, line 2438: might want to make this a list of "reserved" keys for future expansion

aywaldron: yes great idea

README.md
```diff
@@ -35,6 +35,11 @@ Please see the [Redis Talks and Slides](https://github.com/elementary-robotics/a
 
 ### Quickstart
 
+After cloning the repo, clone all submodules with
+```
+git submodule update --init --recursive
```
dpipemazo, line 40: :+1: this is great, thanks!

doc/source/includes/sdk-specification.md
```diff
@@ -1599,6 +1599,253 @@ PERSIST $key
...
+
+### API
+
+| Parameter | Type | Description |
+|-----------|------|-------------|
+| `key` | str | Key at which to store the parameter |
+| `data` | dict | Dictionary containing key-val pairs to store in Redis hash at specified key |
+| `override` | bool | Boolean for whether parameter values may be overriden by future calls to `parameter_write` |
+| `serialization` | str | The method of serialization to use on the parameter fields; defaults to None. |
+| `timeout_ms` | int | How long the parameter should exist in Atom. This sets the expiry timeout in redis. Units in milliseconds. Set to 0 for no timeout, i.e. parameter exists until explicitly deleted. |
```
dpipemazo, line 1651: I think the default timeout on parameters should be infinite unlike references, what do you think? Regardless of choice might be nice to document the default here.

doc/source/includes/sdk-specification.md
```diff
@@ -1599,6 +1599,253 @@ PERSIST $key
...
+
+### Return Value
+
+Tuple of key parameter is stored at, and list of parameter fields written to.
```
dpipemazo, line 1655: Do we need to return and pass around the key? This is what we do on references, but I think that's because we use random strings in the reference keys.

If possible, it would be nice to have an API where we didn't need to track this key and could just `parameter_write("some_key")` and `parameter_delete("some_key")`

doc/source/includes/sdk-specification.md
```diff
@@ -1599,6 +1599,253 @@ PERSIST $key
...
+
+### Spec
+
+1. Make a parameter ID using the following string format:
+```
+key = parameter:$key
+```
+2. Check if the parameter already exists in atom; if so, check its override value. If its override is set to false,
+raise an Exception. Otherwise, continue writing the parameter.
```
dpipemazo, line 1664: :+1:

languages/python/atom/config.py
```diff
@@ -30,6 +30,11 @@
 COMMAND_LIST_COMMAND = "command_list"
 RESERVED_COMMANDS = [COMMAND_LIST_COMMAND, VERSION_COMMAND, HEALTHCHECK_COMMAND]
 
+# Parameters
+OVERRIDE_PARAM_FIELD = "override"
+SERIALIZATION_PARAM_FIELD = "ser"
+RESERVED_PARAM_FIELDS = [OVERRIDE_PARAM_FIELD, SERIALIZATION_PARAM_FIELD]
```
dpipemazo, line 36: :+1:

languages/python/atom/element.py
```diff
@@ -590,15 +597,17 @@ def _metrics_validate_labels(self, labels):
         if "" in labels.values():
             raise AtomError("Metrics labels cannot include empty strings")
 
-    def _make_reference_id(self):
+    def _make_reference_id(self, key=None):
```
dpipemazo, line 600: were the changes here intentional?

languages/python/atom/element.py
```diff
@@ -2223,6 +2232,276 @@ def log(self, level, msg, stdout=True, _pipe=None, redis=False):
...
+        )
+        self._parameter_write_metrics["check"] = self.metrics_create(
+            MetricsLevel.TIMING,
+            "atom:parameter_write",
+            "check",
+            agg_types=["AVG", "MIN", "MAX"],
+        )
+
+    def parameter_write(
+        self, key, data, override=True, serialization=None, timeout_ms=10000
```
dpipemazo, line 2266: Let's set the default to infinite time on parameters -- I see them being used much less frequently than references and for things like system configurations which get written at boot and read for the uptime of the device.

languages/python/atom/element.py
```diff
@@ -2223,6 +2232,276 @@ def log(self, level, msg, stdout=True, _pipe=None, redis=False):
...
+        timeout_ms amount of time.
+
+        If the store already exists under the specified key, the override
+        field will be checked. The fields will be updated if override is true,
+        otherwise an error will be raised.
+
+        Args:
+            key (str): name of parameter to store
+            data (dict): dictionary of data fields to store
+            override (bool, optional): whether or not hash fields can be
```
dpipemazo, line 2282: this is subtle but a **really nice** design I think. As implemented, it seems that the override field will allow for adding **new** fields but not overwriting any existing ones?

languages/python/atom/element.py
```diff
@@ -2223,6 +2232,276 @@ def log(self, level, msg, stdout=True, _pipe=None, redis=False):
...
+
+            if exists[0]:
+                # Check for requested fields
+                for field in data.keys():
+                    _pipe.hget(key, field)
+
+                fields_exist = _pipe.execute()
+                if any(fields_exist):
+                    # Check override setting and raise error if false
+                    _pipe.hget(key, OVERRIDE_PARAM_FIELD)
```
dpipemazo, line 2318: Maybe we should check for the override before checking the keys since it might overall save time? If the override is false we don't have to run all of the `hget`s on the keys

languages/python/atom/element.py
```diff
@@ -2223,6 +2232,276 @@ def log(self, level, msg, stdout=True, _pipe=None, redis=False):
...
+            _pipe = self._release_pipeline(_pipe)
+            self.metrics_timing_end(
+                self._parameter_write_metrics["data"], pipeline=pipeline
+            )
+
+        if not all(response):
+            raise ValueError(f"Failed to create parameter! response {response}")
+
+        # Return key and all fields written to
+        return key, fields
```
dpipemazo, line 2364: I see we're returning the original key -- was there a particular reason to do this?

languages/python/atom/element.py
```diff
@@ -2223,6 +2232,276 @@ def log(self, level, msg, stdout=True, _pipe=None, redis=False):
...
+        return return_data if return_data else None
+
+    def parameter_delete(self, key):
+        """
+        Deletes a parameter and cleans up its memory
+
+        Args:
+            keys (str): Key of parameter to delete from Atom
+        """
+        self.reference_delete(f"parameter:{key}")
```
dpipemazo, line 2476: should we make a helper function to make this string of `parameter:{key}` to keep this DRY?

languages/python/atom/element.py
```diff
@@ -2248,7 +2527,7 @@ def _reference_create_init_metrics(self):
         )
 
     def reference_create(
-        self, *data, serialization=None, serialize=None, timeout_ms=10000
+        self, *data, key=None, serialization=None, serialize=None, timeout_ms=10000
```
dpipemazo, line 2530: I think this might be an improvement, but it might have also slipped in?

dpipemazo (changes requested):
Overall this feature looks great and the API spec is pretty clever. I really like the **override** field applying on a per-key basis.

doc/source/includes/sdk-specification.md
aywaldron: Yeah I was thinking about that too -- if so, would we need to make the element that writes them responsible for deleting them on shutdown? Or do you not see there being any issue with them never being removed explicitly?

doc/source/includes/sdk-specification.md
aywaldron: We don't have to! As written, the API already works like you described. I can remove the key return without having to make any other changes.

languages/python/atom/element.py
aywaldron: I just left in the work I previously did on this ticket which allows users to specify their own reference key since I thought we wanted that enhancement as well. I can remove if we don't need this now that we have parameters?

languages/python/atom/element.py
aywaldron: Sounds good :+1: do we need an explicit removal policy to be documented or just let them sit around?

languages/python/atom/element.py
aywaldron: Correct, if override is false you cannot overwrite existing fields, but you could add brand new ones.

languages/python/atom/element.py
aywaldron: Sure that works. As written, we save time on checking override if no requested keys already exist, but you're right that we'll save more time on average by checking override first since there will be >1 requested key on average.

languages/python/atom/element.py
aywaldron: I'll remove this :)

doc/source/includes/sdk-specification.md
dpipemazo: I imagine the total amount of data here will be negligible and we can document that the creator should clean them up. If we wanted to we could log all parameters created in a dict on the element itself subtly and then clean them up in the destructor, but this might be overkill.

languages/python/atom/element.py
aywaldron: This is the previous version of the work I did on adding user-specified keys to references. Can remove if we don't want this anymore.

doc/source/includes/sdk-specification.md
dpipemazo: Yeah, I think it might be a bit confusing for those familiar with the reference API?

languages/python/atom/element.py
dpipemazo: ah, that's great. Did we update the reference docs for this?

dpipemazo: Definitely still a nice enhancement.

doc/source/includes/sdk-specification.md
aywaldron: Agree, would be a good safeguard but likely overkill as long as `parameter_write` is not misused by any elements. Is redis cleared when we restart nucleus/restart a stack?

languages/python/atom/element.py
aywaldron: Oh good reminder, I didn't update the reference docs -- will add that.

doc/source/includes/sdk-specification.md
dpipemazo: redis is cleared on nucleus restart :+1:

languages/python/atom/element.py
dpipemazo: I think we should maybe just document that they'll live forever by default and the user should be careful to clean them up.

languages/python/atom/element.py
dpipemazo: Yeah, I think it's not clear exactly which use case will be more common but I think the fail case is worse for large numbers of keys and no override than a single extra call. I do tend to agree more keys is likely to be more common as well.

aywaldron:
Thanks for the feedback @dpipemazo - I have pushed the following changes:
- Updated default parameter timeout to be 0 (no timeout)
- Moved override check before requested fields check when parameter already exists
- Added check for matching serialization method when attempting to update parameter (decided not to support switching serialization method after initial parameter write)
- Updated reference API to allow for multiple user-specified keys (previously only worked when creating one reference at a time)
- Updated reference documentation with new user-specified key capability

languages/python/atom/element.py
```diff
@@ -2223,6 +2232,325 @@ def log(self, level, msg, stdout=True, _pipe=None, redis=False):
...
+            _pipe.exists(key)
+            key_exists = _pipe.execute()[0]
+
+            if key_exists:
+                # Check requested serialization is same as existing
+                _pipe.hget(key, SERIALIZATION_PARAM_FIELD)
+                existing_ser = _pipe.execute()[0]
+
+                if existing_ser != serialization.encode():
+                    raise AtomError(
```
dpipemazo, line 2327: :+1: nice!

languages/python/atom/element.py
```diff
@@ -2223,6 +2232,325 @@ def log(self, level, msg, stdout=True, _pipe=None, redis=False):
...
+
+        Args:
+            key (str): Key of a parameter for which we want to update the
+                        timeout
+            timeout_ms (int): Timeout at which we want the key to expire.
+                        Pass <= 0 for no timeout, i.e. never expire (generally
+                        a terrible idea)
+
+        """
+        self.reference_update_timeout_ms(self._make_parameter_key(key), timeout_ms)
```
dpipemazo, line 2541: Similar to `DEL`, does `PEXPIRE` also work on hashes? That's nice.

dpipemazo (approved):
Overall looks great. It's nice that DEL and PEXPIRE work on both keys and hashes to share that logic under the hood.

languages/python/atom/element.py
aywaldron: Yes, so it actually works on Redis keys, and it doesn't matter what data type is stored in the key.

PR #395, opened 2021-01-09
element: add shared counters and sorted sets

dpipemazo:
APIs:
- set
- update (pos/neg compatible)
- get
- delete

Also added a RedisPipeline class similar to the MetricsPipeline
class to clean up some of the logic around getting/releasing
shared pipelines. Didn't apply it throughout the element yet
as this feature should come in before some general cleanup that
needs to be done on the element class.

Also cleaned up some shared logic between references, parameters
and counters as they all use underlying redis key type commands


dpipemazo:
This PR also now includes some dashboard improvements:
1. Dashboard loader code will now poll every 10s for new elements and make new dashboards if it finds them. This is pretty useful if we have elements that are dynamically created/deleted during runtime.
2. Default element dashboard now contains info about references deleted which was a large missing piece
3. Default element dashboard now contains parameter info
4. Default element dashboard now contains counter info including value of counters over time

While doing the above the parameter metrics were tweaked a bit to make the dashboards work a little bit more effectively.

Another small change was removing the use of `KEYS` to scan for elements/dashboards and moving it to use `SCAN` as `KEYS` is not recommended for production use for performance reasons. This has been done in the element as well as in the dashboard code.

Screenshot from 2021-01-09 15-14-01

dpipemazo:
Screenshot from 2021-01-09 16-12-24

We now are logging parameter updates and displaying them with a decently readable timestamp in the metrics dashboards. This is done with a new `metrics_log` API which just writes to a stream in the metrics redis each time a parameter is updated.

dpipemazo:
Screenshot from 2021-01-10 09-40-33

Added in sorted set support including metrics and dashboards as this is pretty useful for handling multi-element/multi-process operations with shared resources (such as global disk space). Can use the sorted set to keep track of file age and when the disk space counter exceeds the max we can atomically have each process pull the oldest file and remove it. This largely is required since Linux doesn't have a good way of keeping track of this sort of thing across a directory with thousands of files without having to recalculate fully or re-scan each time.

languages/python/atom/element.py
```diff
@@ -2944,6 +3059,592 @@ def reference_get_timeout_ms(self, key):
...
+
+    def sorted_set_range(self, set_key, start, end, least=True, withvalues=True):
+        """
+        Read a range of the sorted set
+
+        Args:
+            set_key (string): Name of the sorted set
+            start (int): start index of the read
+            end (int): End index of the read
+            least (bool): Read from least to greatest, else greatest to least
```
aywaldron, line 3481: Is the difference in the response returned from each of these options just the order of the values returned? As with the `pop` function I think `least` is a slightly confusing arg name, maybe something like `descending` vs ascending would be more clear.

languages/python/atom/element.py
```diff
@@ -2944,6 +3059,592 @@ def reference_get_timeout_ms(self, key):
...
+                pipeline=metrics_pipeline,
+            )
+
+    def sorted_set_pop(self, set_key, least=True):
+        """
+        Pop the value from a sorted set. Minium or maximum (min by default)
+
+        Args:
+            set_key (string): Name of the sorted set
+            least (bool): True to pop minimum, False to pop maximum
```
aywaldron, line 3430: I think setting `least=False` in order to pop the max value is slightly confusing/non-intuitive, I wonder if `sorted_set_pop_max` and `sorted_set_pop_min` methods would be a clearer API? or maybe a different arg name could help?

aywaldron (approved):
Lots of great improvements here. Updates to parameter metrics, generalized redis key API, dashboard improvements, etc all look good to me. Couple small comments on the new sorted set API design but nothing major.

languages/python/atom/element.py
dpipemazo: yeah, I agree that `least` is not great. I do prefer an arg as opposed to another function. I wanted it to be `min`, but `min` is already a built-in function in Python. Perhaps it could be `maximum` and be default false, s.t. the only time it's used is with `maximum=True` as opposed to `least=False`?

languages/python/atom/element.py
dpipemazo: Perhaps the same `maximum` API discussed below would be better? Makes sense to name it after the one that will be set/used.

languages/python/atom/element.py
aywaldron: I like `maximum` defaulting to False and only being used with `maximum=True` :+1:

dpipemazo: sounds great, really appreciate this feedback. I'll switch to that before merging.

languages/python/atom/element.py
aywaldron: I think I'm confused on exactly how it works as written, if you change the order through the `least` value do you also need to swap the `start` and `end` values or is `start` always assumed to be less than `end`? If its doing what I think it is, I'd prefer something like a default of `desc=False` which would then only be used when `desc=True`. Unless there is something to be gained by having the args match between this and the `pop` function. But it seems to me the functionality is not the same between the two.

dpipemazo: ah, it actually changes the interpretation of the start and stop. If least is true, index 0 is the least and -1 is the most. If least is false, index 0 is the most and -1 is the least.

PR #400, opened 2021-01-24
Sorted Set Improvements, Add `isort` to formatter, Change `delete` to `unlink`

dpipemazo:
Overall this PR adds some nice things to the sorted set API

- `sorted_set_add` and `sorted_set_pop` now return the cardinality of the set. We were already doing this under the hood, and we have some applications where this is useful to know.
- Added `sorted_set_pop_n` API. This allows us to pop as many as N items from the sorted set and is useful when trying to do batch processing on items from the sorted set.
- `sorted_set_pop` now has blocking API very similar to Python's multiprocessing Queue(). This is useful when having a producer/consumer type architecture based on the sorted set.

-----------------------------------------------------------------------------------------------------------

There's also a fundamental Atom change in here that should be discussed. When implementing the blocking API for `sorted_set_pop` we were seeing the `BZPOP` instructions timing out immediately. Upon reading the [docs](https://redis.io/commands/blpop) for this/similar instructions in Redis to see why this might be happening it was found to be because our `pipeline` objects were all being initialized as **transactional**. This meant that when we called `pipeline.execute()` our instructions were being wrapped in a Multi/Exec that would be guaranteed to all run before another client's commands would be executed. This clearly doesn't play nicely with blocking-type commands (don't want one client putting a blocking command in a pipeline and tying up redis for everyone else). 

As far as I'm aware we don't need the multi/exec functionality anywhere in our spec as we never have two commands dependent on each other and requiring atomic execution. As such, **all pipelines now turn off multi/exec by default**. This allows us to not really change our code and still get the transactional benefit of pipelining. If we do wind up needing it in the future, I'm sure we can find a way to turn multi/exec on for certain calls of pipelines but this has not yet been built in.

In looking at the code it seems we never used pipelines on blocking calls in the past (XREADs and XREADGROUPs all went directly through the redis client) as I imagine we would have had similar issues.


dpipemazo:
This PR still needs doc updates before merging.

dpipemazo:
A few more changes have come in on this PR:

1. Add `isort` to the formatter by default. This cleans up/makes nicer our includes at the top of files and is a much-needed improvement. There is a bug in `isort` when run in a Docker container to reformat that causes it to change the file's owner to `root`. I've documented how to work around this and filed an issue on their repo to track it.
2. Changes redis `delete` calls to `unlink` calls. `del` is a **synchronous** call to free the memory while unlink is **asynchronous** and doesn't actually reclaim the memory. This should help with performance issues. https://redis.io/commands/unlink

PR #404, opened 2021-02-10
atom: add shared queue type around sorted set

### LINKED ISSUE #403: Add shared queue classes around Sorted Sets
#404

dpipemazo:
- Can implement multi-element PRIO and FIFO queues based on the
redis sorted set + pickle() for data for now. Once we get more
language support we can add other serialization methods for the
queue other than pickle()
- These queues are really nice compared to standard multiprocessing
queues in that they allow introspection (can read them without
popping) and can allow for true multi-element support vs processes
spawned from the same element.

A FIFO queue is similar to a stream here, except that by implementing
the FIFO as a prio queue ordered by timestamp, we can forgive slight
out-of-order addition due to multiprocessing races which can be
nice in some instances. The PRIO queue is the major thing in here
that's really nice and there's no real analog to in multiprocess
Python.

Closes #403


dpipemazo:
@slaigai this is now rebased on the ARM executor build and is working; let's bring over the latest version of the queues with the type hints and work to get this in. I had also scaled down the tests a bit originally (max throughput 100 instead of 1000), we should put it back to 1000 now that we don't have the ARM perf issues. Thanks!

languages/python/atom/config.py
```diff
@@ -56,6 +56,45 @@
     (3600000, 2592000000),
 ]
 
+# Queue metrics
+METRICS_QUEUE_TYPE = "queue"
```
dpipemazo, line 60: What do you think about moving these into a  "metrics.py" file or similar? We've been overloading `config.py` for a while with both config and constants.

languages/python/atom/element.py
```diff
@@ -80,15 +80,22 @@
 DuplicatePolicy = Literal["block", "first", "last", "min", "max"]
 
 
-class HandlerDict(TypedDict, total=False):
+class HandlerDictBase(TypedDict):
```
dpipemazo, line 83: I think the handler part here might be leaking from the original implementation; do we need this?

dpipemazo (commented):
Overall looks great. We could do it in this PR or in follow-on PRs, but let's add:
1. Some docs on the queue, its benefits and how to use it
2. Some docs on the dashboards for the queue and how to use them. 
3. A test case/example using the queue in Python, which can help to make/show dashboards for (2)

Overall this is great! Thanks for getting the types in here and pushing this forward

languages/python/atom/config.py
slaigai: Could you elaborate on what the difference is between "config" and "constants"? It's nice having one place to find all of these constants but I also see some benefits for placing constants adjacent to the code they're primarily used in (e.g. moving metrics constants to metrics.py like you said).

slaigai:
> Overall looks great. We could do it in this PR or in follow-on PRs, but let's add:
> 
> 1. Some docs on the queue, its benefits and how to use it
> 2. Some docs on the dashboards for the queue and how to use them.
> 3. A test case/example using the queue in Python, which can help to make/show dashboards for (2)
> 
> Overall this is great! Thanks for getting the types in here and pushing this forward

@dpipemazo Good ideas, we should definitely add these docs. In general, I think we should add these in a separate smaller PR for higher PR review quality.

I'd be more than happy to take that ticket as it would be an opportunity to learn these queues inside and out; however, it might be easier/better for you to put these together than if I did it. What do you think about that?

languages/python/atom/config.py
dpipemazo: Yeah, I guess I see "config" as "things users/us might want to change" and constants as "things we don't want to hardcode but users likely shouldn't be changing". It's admittedly a vague line.

dpipemazo:
> I'd be more than happy to take that ticket as it would be an opportunity to learn these queues inside and out; however, it might be easier/better for you to put these together than if I did it. What do you think about that?

I think it would be awesome if you could take it as an opportunity to become more familiar with the queues and be able to teach the rest of us/the team :smile:

languages/python/atom/queue.py
```diff
@@ -0,0 +1,750 @@
+from __future__ import annotations
```
kylebebak, line 1: This looks very similar to `vision/utils/queue.py` in VP, so I figure it works the same as the queues in that module =)

languages/python/atom/element.py
kylebebak, line 83: So this is the way to make some keys required (`handler` and `serialization`) and some not (`deserialize`)

Not the world's prettiest syntax, but this is much better than making all 3 optional when 2 of them really aren't optional!

languages/python/atom/serialization.py
```diff
@@ -6,7 +6,7 @@
 
 import numpy as np
 import pyarrow as pa
-from msgpack import packb, unpackb
+from msgpack import packb, unpackb  # type: ignore
```
kylebebak, line 9: You could create a stubs file for these methods (I did in FastAPI)

https://github.com/elementary-robotics/cloud-fastapi/blob/ce2129dc83574b1b60e10ef3137882ad3327a48f/stubs/msgpack/__init__.pyi#L1-L1

And then you'd have to put something like this in `pyrightconfig.json`:

```
"stubPath": "./stubs",
```

To be clear, **IMO it's not worth it** -- we only use these methods in `serialization.Msgpack`, and the type signatures of the methods on this class don't bleed into the rest of the code base.

The only thing I might do is something like this

```
class Msgpack(GenericSerializationMethod):
    """
    Class containing msgpack serialization and deserialization functions.
    """

    @classmethod
    def serialize(cls, data) -> bytes:
        return packb(data, use_bin_type=True)

    @classmethod
    def deserialize(cls, data: bytes):
        return unpackb(data, raw=False)
```

kylebebak (approved):
Looks great =)

As I read through this I thought it might be good to add type annotations to the test code, and run the test code through `pyright` in CI as well.

The test code is effectively 3.5k lines of atom client code, and it would help catch issues in the types before we catch them downstream in other repos, like vision processing.

For example, in `test_queue_fifo`, you could do:

```
test_q: AtomFIFOQueue[int] = AtomFIFOQueue("some_queue", element, max_len=max_len)
```

...and verify that the generic typing for the queues actually works.

If you think it's a good idea, we should make a separate issue for this =)

BTW, this is something we do in FastAPI: https://github.com/elementary-robotics/cloud-fastapi/blob/ce2129dc83574b1b60e10ef3137882ad3327a48f/src/test_main.py#L1-L1

languages/python/atom/queue.py
dpipemazo: The idea is to move these from the internal repo we currently have them in into the open-source so that they can be shared/utilized in other applications/elements

slaigai:
> As I read through this I thought it might be good to add type annotations to the test code, and run the test code through pyright in CI as well.

I agree, I'll open a separate ticket for this.

languages/python/atom/config.py
slaigai: That makes perfect sense to me now, though I can't tell which things are config vs constants. Case in point haha.

languages/python/atom/serialization.py
slaigai: The idea of adding `bytes` to those method signatures makes total sense. I tried adding these without type stubs but pyright still complains about the return type of `packb` being `Unknown | None`. Type stubs might be needed in order to hint the return type here.

PR #406, opened 2021-02-18
Remove lazycontract (not used)

### LINKED ISSUE #405: Get rid of lazycontract
Code in `languages/python/atom/contracts.py` is no longer used in Atom code base, except for one test case in `languages/python/tests/test_atom.py`.

This means we can also remove `lazycontract` dep.

Worth noting that `spec-robot` and `element-realsense` depend on these contracts:

https://github.com/elementary-robotics/spec-robot/blob/43235d98e3330a38c3f5ea581f8ec5f85e0717c9/spec_main/contracts.py#L6-L6

But spec-robot is pinned to `ARG ATOM_IMAGE=elementaryrobotics/atom:development-v2.0.0-beta-opencv-amd64`, and realsense is pinned to `elementaryrobotics/atom:v1.0.1`.

We can remove these contracts from atom and not use them going forward, but spec-robot and realsense will still work if we need them to. If we need to use spec-robot or realsense again, it would also be trivial to get rid of these deps.

kylebebak:
Closes #405 

- Nothing we currently use in Atom or libraries that depend on atom uses these contracts (more discussion in #405)
- ~~Also prepares `setup.py` and other files needed to publish atom to PyPI~~

Second item will be handled in a separate PR.

~~Once we publish atom as a library, it will be easy to include it as a dev dependency in any repo that depends on atom, which means we can get type checking and auto-completion on any calls to atom methods~~

https://pypi.org/search/?q=atom

We just need to pick a name. The package name "atom" is taken. "elementary-atom" and "atom-elementary" are not


dpipemazo (approved):
:+1: looks good. Just re-running the failed build job to see if it was a fluke (should be, not sure what could have broken just that one build)

languages/python/setup.py
```diff
@@ -10,6 +10,10 @@
 setuptools.setup(
     name="atom",
     packages=["atom"],
+    description="Python bindings for the Elementary Robotics Atom SDK",
+    url="https://github.com/elementary-robotics/atom",
+    author="Elementary Robotics",
+    install_requires=["msgpack==0.6.2"],
```
dpipemazo, line 16: Don't we also need redis-py, redistimeseries-py, etc?

dpipemazo (changes requested):
@kylebebak build failure doesn't seem to be a fluke; not sure why it's picking up python2 as the default in that base image but we need to look into it.

languages/python/setup.py
kylebebak: You're right -- I didn't see these in `requirements.txt`, but see they're in `Dockerfile-base` and `Dockerfile`.

kylebebak:
Might be related to `setup.py`, `install_requires=["msgpack==0.6.2"]`. I'll have a look

languages/python/setup.py
dpipemazo: I think we also need numpy, arrow, etc. Overall it seems like we're working on multiple features here and I'd recommend scoping this down to removing `lazycontract` and then getting a separate issue to get us on Pip. Grouping multiple things into one PR is likely not the best idea.

kylebebak: @dpipemazo 

Agreed, I thought this would be simple and quick, and it wasn't, but even if it were putting these both into the same PR isn't a good idea.

I've scoped it down to just removing `lazycontract`

dpipemazo:
@kylebebak turns out it was in Redis Time Series: https://github.com/RedisTimeSeries/RedisTimeSeries/issues/660

I'm fixing it in the PR to bring all of our builds up to Python3.7, and then we can rebase this atop that.

dpipemazo (changes requested):
@kylebebak I think it's still in the `requirements.txt`

dpipemazo (approved):

PR #408, opened 2021-02-24
Add type hints to atom python

### LINKED ISSUE #407: Add type annotations to `languages/python/atom/element.py`
This is actually where the type hint value:effort ratio is highest, because pretty much everything running on the "camera side" depends on Atom.

We don't need all the code in this repo to be annotated, just the method/function signatures, so that anything that depends on Atom can use those type hints (e.g. vision-processing, to which we're adding type hints now, https://github.com/elementary-robotics/vision-processing/issues/126).

Adding type hints to method/function signatures would require minimal effort.

### LINKED ISSUE #418: Switch to pip installing redistimeseries
No longer need to build from source

slaigai:
Changes:
- add type hints to atom (mostly in element.py)
- add type checker (pyright) to test container and include in build checks
- udpate redistimeseries in atom to 1.4.3 using pip to install (stable latest and greatest). This resolves import errors from pyright because redistimeseries used to be a binary built from source. Now pyright can resolve imports correctly.
- update logic to resolve type errors (see changes to tests as well)

Closes #407 , closes #418


dpipemazo:
@slaigai may be a bit of a naive question but how do we know that the type hints added are correct? Are we running `pyright` or similar on the test cases in CI to ensure/enforce/validate the type hinting in the APIs?

slaigai:
> @slaigai may be a bit of a naive question but how do we know that the type hints added are correct? Are we running `pyright` or similar on the test cases in CI to ensure/enforce/validate the type hinting in the APIs?

Oh, right. Not a naive question at all. We still need to figure out how we're going to set up the type checker per #413 . Meanwhile, I will expand the scope of this PR to include the following:
- add type hints
- add type checker to atom repo CI baked into test container. follow up with transition after #413  to replace this type checker
- fix type errors raised by type checker

Dockerfile
```diff
@@ -220,6 +217,12 @@ RUN apt-get install -y --no-install-recommends valgrind
 ADD ./languages/python/requirements-test.txt .
 RUN pip3 install --no-cache-dir -r requirements-test.txt
 
+# Install pyright
```
kylebebak, line 220: Cool, same recipe for adding type checking to test container in VP

languages/python/atom/messages.py
```diff
@@ -7,7 +12,17 @@
 ENTRY_RESERVED_KEYS = "ser"
 
 
-def format_redis_py(data):
+@overload
```
kylebebak, line 15: Nice use of overload. It also makes it more obvious that this function signature is a little weird, which is a nice benefit of adding type hints!

languages/python/atom/messages.py
```diff
@@ -44,7 +59,12 @@ def __init__(self, element, cmd, data):
 
 class Response:
     def __init__(
-        self, data="", err_code=0, err_str="", serialization=None, serialize=None
+        self,
+        data="",
+        err_code: int = 0,
+        err_str: str = "",
+        serialization: Optional[str] = None,
+        serialize: Optional[bool] = None,
```
kylebebak, line 67: Tracing the code back to the `serialization.py` module, I'm pretty sure `serialization` here, and probably everywhere else, must be one of `"msgpack", "arrow", "none"`

Let's use `Literal`!

languages/python/atom/serialization.py
```diff
@@ -102,15 +105,15 @@ def print_values(cls):
...
+def is_valid_serialization(method: Optional[str]) -> bool:
     """
     Checks serialization method string against available serialization options.
     Returns True/False if method is valid/invalid.
     """
     return (method in Serializations.__members__) or (method is None)
 
 
-def serialize(data, method="none"):
+def serialize(data, method: Optional[str] = "none"):
```
kylebebak, line 116: Let's also use `Literal` and make this one of `"msgpack", "arrow", "none"`

languages/python/atom/serialization.py
```diff
@@ -135,7 +138,7 @@ def serialize(data, method="none"):
     return Serializations[method].value.serialize(data)
 
 
-def deserialize(data, method="msgpack"):
+def deserialize(data, method: Optional[str] = "msgpack"):
```
kylebebak, line 141: Ditto

languages/python/atom/element.py
```diff
@@ -187,51 +195,59 @@ def __init__(
 
         self.name = name
         self.host = uname().nodename
-        self.handler_map = {}
-        self.timeouts = {}
+        self.handler_map: dict[str, Any] = {}
```
kylebebak, line 198: Can't be sure this is always true, but it looks like the values in `self.handler_map` have the following shape

```py
        self.handler_map[HEALTHCHECK_COMMAND] = {
            "handler": handler,
            "serialization": None,
        }
```

We should use [TypedDict](https://www.python.org/dev/peps/pep-0589/) here. Something like:

```
class Handler(TypedDict):
    handler: Callable
    serialization: Optional[Literal[...]]
```

Then we'd have `self.handler_map: dict[str, Handler]`

languages/python/atom/element.py
```diff
@@ -1854,13 +1890,13 @@ def command_send(
...
     def entry_read_loop(
         self,
-        stream_handlers,
-        n_loops=None,
-        timeout=MAX_BLOCK,
-        serialization=None,
-        force_serialization=False,
-        deserialize=None,
-    ):
+        stream_handlers: Sequence[StreamHandler],
```
kylebebak, line 1893: Awesome =)

Also, I had a look at `StreamHandler`, and noticed that it has some validation code:

```
        if not isinstance(element, str):
            raise TypeError("element must be a str")
        if not isinstance(stream, str):
            raise TypeError("stream must be a str")
        if not callable(handler):
            raise TypeError("handler must be a function")
```

Given that the signature has type hints, and we're enabling type checking in this repo, I think this code, and code like this, is probably redundant an undesirable (I gave a bit of similar feedback in the VP types PR)

languages/python/atom/element.py
```diff
@@ -2192,12 +2230,12 @@ def _entry_write_init_metrics(self, stream_name):
...
     def entry_write(
         self,
-        stream_name,
-        field_data_map,
-        maxlen=STREAM_LEN,
-        serialization=None,
-        serialize=None,
-    ):
+        stream_name: str,
+        field_data_map: dict,
```
kylebebak, line 2234: Can this be any more explicit, e.g. `dict[str, Any]`?

languages/python/atom/element.py
```diff
@@ -2686,9 +2746,18 @@ def reference_create(
...
-        keys = [keys] if type(keys) is not list else keys
-        if len(data) != len(keys):
+        if keys is None:
+            keys_list = [None] * len(data)
+        elif type(keys) is not list and type(keys) is str:
+            keys_list = [keys]
+        elif type(keys) is list and all(isinstance(k, str) for k in keys):
+            keys_list = keys
+        else:
+            raise TypeError(
```
kylebebak, line 2756: Very nice =)

This is a good example fo where raising this type error is necessary, even though **within this repo** the type checker would protect us from passing in an invalid type.

Methods exposed on `Element`, such as `reference_create`, are used and will be used in edge repos without type checking, which means we can't assume client code in other repos won't pass in invalid types for arguments like `keys`

kylebebak (changes requested):
Looks great! Just some general feedback and a few suggestions :

- remove validation code if it's unnecessary (e.g. in `Acknowledge`, `Log`, `StreamHandler`) in `messages.py`
  + all this validation would be unnecessary if we were using `pyright` in other repos that use these classes; it's __unnecessary now__ if other repos don't use these classes, and they're only used internally; for example, the type checker simply won't allow us to pass in an `int` for an `element` argument of type `str`
- tighten up types where possible in containers (especially dicts)
- use `TypedDict` where appropriate
- use literal where appropriate, e.g. wherever we have a `serialization` argument

@slaigai Also, I think we should add documentation to README on how to run type checking in this repo, similar to what you did in VP.

-----

And one other thing that should probably be a separate issue, and that dovetails with publishing atom on PyPI. I created this issue here:

slaigai:
> use literal where appropriate, e.g. wherever we have a `serialization` argument

@kylebebak good suggestion. I've created a type alias in `serialization.py` that's reused across atom. We can import type aliases just like any other object from the `serialization` module. As you might have already figured out, this encourages code reuse and makes sure that there is exactly one place to edit when modifying the list of valid serialization methods.

languages/python/atom/element.py
slaigai: Thanks for calling this out, you gave useful insights here about redundant validation code! While I would agree with removing type validations for private functions, I have concerns about removing type checks for the public API (primarily those in messages.py). These may be imported into contexts where type checking is not performed as you mentioned in your review comment.

> if we think a class might be used by other repos, having this validation code might be useful, because we don't know if those repos are using pyright; most of them aren't

From a risk/benefit perspective, I think we'd incur more risk (e.g. hard-to-trace bugs, unusual error modes from incorrect usage) by removing these checks than any performance or maintainability gains. What do you think @kylebebak ?

languages/python/atom/element.py
```diff
@@ -77,6 +72,25 @@
...
 )
+from redis.client import Pipeline
+from redistimeseries.client import Client as RedisTimeSeries
+from redistimeseries.client import Pipeline as RedisTimeSeriesPipeline
+from typing_extensions import Literal, TypedDict
+
+DuplicatePolicy = Literal["block", "first", "last", "min", "max"]
+
+
+class HandlerDict(TypedDict, total=False):
```
kylebebak, line 83: Nice use of `total=False` here so that deserialize isn't a required key (which is different from being `Optional`)

languages/python/atom/element.py
```diff
@@ -77,6 +72,25 @@
     StreamHandler,
     format_redis_py,
 )
+from redis.client import Pipeline
+from redistimeseries.client import Client as RedisTimeSeries
+from redistimeseries.client import Pipeline as RedisTimeSeriesPipeline
+from typing_extensions import Literal, TypedDict
+
+DuplicatePolicy = Literal["block", "first", "last", "min", "max"]
```
kylebebak, line 80: String literal union types =)

languages/python/atom/serialization.py
```diff
@@ -7,6 +7,9 @@
 import numpy as np
 import pyarrow as pa
 from msgpack import packb, unpackb
+from typing_extensions import Literal
+
+SerializationMethod = Literal["msgpack", "arrow", "none"]
```
kylebebak, line 12: This is used in like 20 places, this is excellent

languages/python/atom/element.py
kylebebak: I think you're right. We talked about this a bit, and concluded that code in Atom should be written defensively, because repos that depend on atom might call code in this repo, and they might not be using a checker

Code in other repos, like vision processing, should not have defensive "validation" code if they have type hints

kylebebak (approved):
LGTM, this is ready to go!

PR #411, opened 2021-02-27
Add type checking to formatter

### LINKED ISSUE #410: Add type checking to formatter
If type hints are available, a type checker can check for errors and more. We want type checking disabled by default, enabled by environment variable

slaigai:
- Add pyright type checker to formatter
- Update instructions in README for running formatter with type checking
- Added environment variable `CHECK_ISORT` (default on) while linting
- Update formatter image to latest and greatest in order to install pyright
  - Pyright is installed using npm. `alpine/flake8:3.7.0` would install npm but would error upon `npm install -g pyright` due to a bug. Bugfixes that unblock pyright installation are available in `alpine/flake8:3.8.4`. 

Closes #410

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,32 @@ docker run -e BLACK_EXCLUDE="" -e FLAKE8_EXCLUDE="" --mount src=$(pwd),target=/c
 
 After running this, check the return code with `echo $?` to see if the checks passed. If the return code is 0, then the checks passed, if nonzero then there were formatting inconsistencies. The logs will also be indicative.
 
+To check your code using a type checker:
+```
+docker run -e BLACK_EXCLUDE="" -e FLAKE8_EXCLUDE="" -e CHECK_PYRIGHT="y" \
+  --mount src=$(pwd),target=/code,type=bind \
+  --mount src=<virtualenv_dir>,target=/.venvs,type=bind,readonly \
+  elementary/formatter .
+```
+
+This runs the pyright type checker. A file named `pyrightconfig.json` must exist
+in the current directory (example below). Replace `<virtualenv_dir>` in the
+docker command with the path to a directory containing one or more
+subdirectories, each of which contains a virtual environment. This enables
+pyright to discover packages while type checking. Check the [pyright
+documentation](https://github.com/microsoft/pyright/blob/master/docs/configuration.md)
+for configuration details.
+```json
+{
+    "exclude": [
+        "**/__pycache__",
+    ],
+    "venvPath": "/.venvs",
+    "pythonVersion": "3.7",
+    "pythonPlatform": "Linux"
+}
+```
+
 To reformat your code:
 ```
 docker run -e DO_FORMAT=y -e BLACK_EXCLUDE="" -e FLAKE8_EXCLUDE="" --mount src=$(pwd),target=/code,type=bind elementaryrobotics/formatter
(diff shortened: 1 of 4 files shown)


utilities/formatting/Dockerfile
```diff
@@ -1,11 +1,12 @@
-FROM alpine/flake8:3.7.0
+FROM alpine/flake8:3.8.4
```
kylebebak, line 1: Very nice, installing `npm` and `pyright` =)

README.md
```diff
@@ -131,6 +131,32 @@ docker run -e BLACK_EXCLUDE="" -e FLAKE8_EXCLUDE="" --mount src=$(pwd),target=/c
...
+```
+
+This runs the pyright type checker. A file named `pyrightconfig.json` must exist
+in the current directory (example below). Replace `<virtualenv_dir>` in the
+docker command with the path to a directory containing one or more
+subdirectories, each of which contains a virtual environment. This enables
+pyright to discover packages while type checking. Check the [pyright
+documentation](https://github.com/microsoft/pyright/blob/master/docs/configuration.md)
+for configuration details.
+```json
```
kylebebak, line 149: Ok, so each dev can have their own `<virtualenv_dir>` that they pass into the `docker run` call, which mounts the virtual env dir into `/.venvs` in the formatter container.

This makes a lot of sense, but I'm not sure the `pyrightconfig.json` file below should be in version control, in this repo or any other repo with Python code.

That file has a hard-coded `venvPath`, which is perfect for running `pyright` in the formatter container, but which isn't going to match the path to the virtual env on a given dev's host machine. Devs will want to run `pyright` on their host machines, which means each devs will want their own `pyrightconfig.json` on their host machines as well.

This is really easy to solve; we can bake a `pyrightconfig.json` file into the formatter container without having one on the host machine.

kylebebak (changes requested):
Checking types in the formatter might not make sense. The formatter isn't just used in development, it's used on the CI server. I had a look at the `check_formatting` job in the Atom orb: https://github.com/elementary-robotics/atom/blob/f7d0dc7ea372e2c6ae30e8c2c1f210542e90d8bc/.circleci/atom.yml#L553-L553

We can use this job in our CI server for any Python repo. The job launches the formatting container, checks out the repo code (e.g. the `vision-processing` code) into `/root/project`, and runs the formatter checks on the code.

Our current "formatter" checks don't require Python deps to be installed to run. `isort`, `black`, and `flake8` don't analyze deps. With type checking, however, whether we use `pyright` or `mypy`, __we need to install Python deps to check types__.

I see two options:

- we teach the formatter how to install Python deps in a way that can be standardized across repos
- we don't do type checking with the formatter; any repo that wants CI type checking does type checking in the "application" container (this is what FastAPI does)

Here is how FastAPI does it:

https://github.com/elementary-robotics/cloud-fastapi/blob/6dd67a8cc64928d9d96c72330bc0795a7d8d7f30/.circleci/config.yml#L37-L37

@dpipemazo

dpipemazo:
@kylebebak it's a good point that we need Python deps to do type-checking. I had not considered that when I asked @slaigai to do it in the atom formatter. 

It is probably best to do it in the repo itself :+1: . @slaigai we can close this out, thanks for getting it up so quickly!

slaigai:
> @kylebebak it's a good point that we need Python deps to do type-checking. I had not considered that when I asked @slaigai to do it in the atom formatter.
> 
> It is probably best to do it in the repo itself  . @slaigai we can close this out, thanks for getting it up so quickly!

I agree, I grappled with this while implementing the changes in this PR. Happy to have learned something that we can take to the other repos!

PR #414, opened 2021-03-04
409/python37

dpipemazo:
This PR primarily ensures that we have Python 3.7 installed in all Atom containers. The primary issue here was that the build of ubuntu shipped in the Nvidia bases we use for the cuda + opengl builds didn't have Python 3.7 as an automatic installation target and instead was installing 3.6.9.

We have done the following:
1. Updated the `Dockerfile-base` to add the `universe` ubuntu repo target (or fail gracefully on this step if it's a non-ubuntu build). This enables installing `python3.7` from `apt` for **all** base images so that the flow can proceed identically for the rest of them.
2. Set `python3.7` as the active python via a symlink after installing it. We're not trying to make sure that 3.7 is the **only** python for now (as this is a container size improvement we can work on later), just that it's there and the primary install target
3. Fixes an intermittent build issue in RedisTimeSeries due to trying to use Python2 for its installation scripts. We now install RTS with Python3. This solves an intermittent bug we've been seeing in some builds, seen in #406 .
4. Adds a proper `requirements.txt` file for the Atom python package. This was obtained by running `pip freeze` on our most recent `latest` build and checking it in. This was both necessary to install all additional requirements needed for Atom and also will make it easier for people to install from PyPi without compiling from source. Since we choose to compile from source, `pip install` when frozen to exact matches, will find existing versions and skip most of it, but for a user who didn't install from source they will be able to install all dependencies via pip.
5. Adds a proper `requirements.txt` file for test dependencies. Currently just `pytest` but also a good thing to have
6. Removes the `redis-py` submodule. We originally had this because @Cody-G had made a custom change to allow it to accept a `memoryview` that we were waiting to be accepted upstream. It was accepted in `3.5.0` (nice work Cody!) so we can go back to installing it via pip. 
7. Adds a test in our CI system to ensure we have a form of Python 3.7 installed in all images. We will now enforce this explicitly as we will depend on it moving forward for our type checking.

Note that with this upgrade path for Python 3.7 we have not changed any of the base images we're building from, so the CUDA and cudnn versions are identical. This is a less risky path and will allow us to upgrade those independently in the future if/when we would like to do so.


slaigai (approved):
Looks good to me. Nice job!

slaigai (approved):
Looks good to me. Nice job!

PR #426, opened 2021-03-11
412/publish on pypi

### LINKED ISSUE #412: Ship Atom on PyPI
Will want to ship a version of Atom on PyPi that has the proper dependencies.

kylebebak:
Closes #412


slaigai:
@kylebebak for my understanding, could you share what you considered when removing `version.py` and switching to manual edits? I'm wondering whether it would be better replace it with a simpler `version.py` that can be used by the 3 python files that depend on it (`setup.py`, `setup_local.py`, and `atom/config.py`).

kylebebak:
Hey @slaigai 

Dan basically explained that `version.py` is buggy and we should get rid of it. It would be nice to build a new, simpler `version.py`, but I think that should go in a separate issue.

Also, I created a `CommandHandler` type that ensures these handlers always return a `Response` object (they have to do this)

languages/python/atom/element.py
```diff
@@ -78,6 +78,7 @@
 from typing_extensions import Literal, TypedDict
 
 DuplicatePolicy = Literal["block", "first", "last", "min", "max"]
+CommandHandler = Callable[..., Response]
```
slaigai, line 81: I learned something new! Great add here

languages/python/atom/element.py
```diff
@@ -4315,6 +4316,9 @@ def metrics_log(
         else:
             _pipe = pipeline
 
+        if _pipe is None:
```
slaigai, line 4319: as it should be :+1:

slaigai (approved):
great work! LGTM

PR #436, opened 2021-03-16
add peek to queues

### LINKED ISSUE #427: Add `read` for new queues
Once #404 is merged, there's one API that would be great to have: the ability to read from the queue without consuming (introspection). This is a huge advantage to the redis sorted set over the Python Multiprocessing Queue.

It should be fairly straightforward to implement using a `ZRANGE` or similar command and an API similar to `get_n`. 

What do you think we should call it? `read` is OK but `inspect` might be better? Open to ideas on this.

slaigai:
Changes:
- Add peek functionality to read at most N items from queues without consuming items.
- Add tests for peek
- Bugfix: AtomFIFOQueue with `max_highest_prio=True` did not return items in FIFO order. This argument is now silently ignored to maintain compatibility with the AtomPRIOQueue interface.

Closes #427


languages/python/atom/element.py
```diff
@@ -3800,10 +3800,9 @@ def sorted_set_range(
             # Add to the sorted set
             self.metrics_timing_start(self._sorted_set_metrics[set_key]["range"])
 
-            if not maximum:
-                redis_pipeline.zrange(redis_key, start, end, withscores=withvalues)
-            else:
-                redis_pipeline.zrevrange(redis_key, start, end, withscores=withvalues)
+            redis_pipeline.zrange(
```
dpipemazo, line 3803: This is not the same as `zrange` vs `zrevrange`. We want to not only display the results in ascending/descending order, but we want to read from the queue in ascending/descending order of priority. Some unit test cases should show/display this:

Take a Queue with 3 items, with priorities 1, 2, 3.

- `peek` of `n=1` should return `1` on a prio queue that's operating on minimum prio and `3` on a prio queue operating on maximum prio. 
- `peek` of `n=2` should return `[1, 2]` on a min prio queue, and `[3, 2]` on a max prio queue.

languages/python/tests/test_queue.py
```diff
@@ -468,3 +468,54 @@ def test_queue_prio_greatest(self, nucleus_redis, element, max_len):
...
+
+    @pytest.mark.parametrize("n", [-1, 0, 1, 2, 3, 4])
+    @pytest.mark.parametrize("num_items", [2, 4])
+    @pytest.mark.parametrize("max_len", [3])
+    @pytest.mark.parametrize("queue_type", QUEUE_TYPES)
+    def test_queue_peek_nonempty(
+        self, nucleus_redis, element, queue_type, n, num_items, max_len
+    ):
+        """
+        Test reading items from a non-empty queue does not consume items. Test
```
dpipemazo, line 480: We should check the values themselves here; it's not enough just to check the length and can/will help catch implementation errors.

dpipemazo (changes requested):

languages/python/atom/element.py
slaigai: Ahh, that makes sense. Thanks for clearing that up

languages/python/tests/test_queue.py
slaigai: As expected, this caught a few extra implementation errors!

dpipemazo (approved):
:+1: looking great, thank you!

PR #439, opened 2021-03-16
438/improve type signature command send

### LINKED ISSUE #438: Improve type signature for command_send
It returns `Dict[str, Any]` now, but we can have it return a typed dict that's more descriptive. This is fairly valuable because this is one of the atom functions that's used in a number of different repos

kylebebak:
Closes #438


languages/python/atom/element.py
```diff
@@ -114,6 +125,13 @@ class CommandHandlerMap(CommandHandlerMapBase, total=False):
 ATOM_LOG_FILE_SIZE = int(os.getenv("ATOM_LOG_FILE_SIZE", LOG_DEFAULT_FILE_SIZE))
 
 
+def get_response_dict(response: Response):
```
slaigai, line 128: nice :+1: rock solid in pyright

slaigai (approved):
solid work!

PR #440, opened 2021-03-18
430/tighten up types

### LINKED ISSUE #430: Tighten up type checks in atom
## Problem

By default, `pyright` runs with `"typeCheckingMode": "basic"`, which misses a lot of things. Most importantly, it doesn't prevent accessing attributes on things that might be `None`, which is one of the most important things we can get from type checking

https://github.com/microsoft/pyright/blob/master/docs/configuration.md

If we make it `"strict"` it catches just about everything, but all the `reportUnknown...` rules are overly aggressive, or at least they are right now in the atom repo

I had a look through the rules, and think a good compromise is to use `"basic"` type checking, but with the `reportOptional...` rules enabled

```
{
  "useLibraryCodeForTypes": true,
  "typeCheckingMode": "basic",
  "reportOptionalSubscript": "error",
  "reportOptionalMemberAccess": "error",
  "reportOptionalCall": "error",
  "reportOptionalIterable": "error",
  "reportOptionalContextManager": "error",
  "reportOptionalOperand": "error",
  "strictListInference": true,
  "strictDictionaryInference": true
}
```

Turning this on locally helped me catch some possible bugs in https://github.com/elementary-robotics/atom/pull/426

But there are still about 40 of these errors remaining in `element.py`

```
/Users/kylebebak/Code/Work/ElementaryRobotics/atom/languages/python/atom/element.py

1619:9 - error: Object of type "None" is not subscriptable (reportOptionalSubscript)
1629:9 - error: Object of type "None" is not subscriptable (reportOptionalSubscript)
...
```

## Solutions

- [ ] mount a `pyrightconfig-ci.json` with these rules into the container, and fix all the "access stuff on `None`" errors
- [ ] modify type checker to use `pyright -p /path/to/pyrightconfig-ci.json`

Then devs can create their own `pyrightconfig.json` based on `pyrightconfig-ci.json` on the host machine, but they can set their own values for `venv` and `venvPath` to point to the right virtual env on the host machine. This file should be gitignored

kylebebak:
Closes #430 

- bumps to version `2.7.1` (we can cut tag and publish release on PyPI once we merge this)

**Most importantly, this actually runs type checks on the source code**. We weren't doing this previously, because the `Dockerfile` wasn't copying in `atom` source code to `/atom/languages/python/atom` for the `test` stage. See this build from a few days ago:

https://app.circleci.com/pipelines/github/elementary-robotics/atom/2037/workflows/9ef6bd4c-ef57-422d-9900-9e961e48a3de/jobs/15756

Screen Shot 2021-03-18 at 8 40 46 AM

We're running tests on 0 files. `pyright` exits with a successful status code, but it doesn't actually run any tests!

Here's how things look in the new build:

Screen Shot 2021-03-18 at 9 01 51 AM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -295,7 +295,7 @@ commands:
           command: docker exec -it -w /atom/languages/python/tests << parameters.container >> python -V 2>&1 | grep "3.7"
       - run:
           name: Python type check
-          command: docker exec -it << parameters.container >> pyright --lib /atom/languages/python/atom
+          command: docker exec -it << parameters.container >> pyright -p pyrightconfig-ci.json /atom/languages/python/atom
       - run:
           name: Python tests
           command: docker exec -it -w /atom/languages/python/tests << parameters.container >> pytest -vv --durations=0 --capture=tee-sys
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -295,7 +295,7 @@ commands:
           command: docker exec -it -w /atom/languages/python/tests << parameters.container >> python -V 2>&1 | grep "3.7"
       - run:
           name: Python type check
-          command: docker exec -it << parameters.container >> pyright -p pyrightconfig-ci.json /atom/languages/python/atom
+          command: docker exec -it << parameters.container >> bash -c "cd /atom/languages/python && pyright -p pyrightconfig-ci.json atom"
       - run:
           name: Python tests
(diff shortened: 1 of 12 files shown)


slaigai (approved):
Looks good! Doesn't really matter whether `typing.Dict` gets swapped out now or later so that's up to you.

PR #444, opened 2021-04-02
clear handlers when creating logger

### LINKED ISSUE #442: Fix duplicate logging
There's a weird bug where logs may be duplicated if an element reuses an element name that has been previously used within the same program. This is because `logger = logging.getLogger(self.name)` in [element.py](https://github.com/elementary-robotics/atom/blob/latest/languages/python/atom/element.py#L302) will [return the same logger instance when called with the same name](https://docs.python.org/3/library/logging.html#logging.getLogger).

This causes [a new rotating file handler is added to the logger](https://github.com/elementary-robotics/atom/blob/latest/languages/python/atom/element.py#L315) every time the element name is reused, leading to duplicate logs.

There's a [quick and easy fix for this](https://stackoverflow.com/a/44049484) that involves clearing the handlers if there are any upon init:
```python
if (logger.hasHandlers()):
    logger.handlers.clear()

logger.addHandler(handler)
```

slaigai:
Closes #442

### DIFF
--- a/languages/python/atom/element.py
+++ b/languages/python/atom/element.py
@@ -300,6 +300,15 @@ class Element:
         # Set up logger
         #
         logger = logging.getLogger(self.name)
+
+        # Ensure only one handler is added to avoid duplicate logs. Calling
+        # getLogger with the same name as a previous element will yield the same
+        # logger instance as before, which already has a handler. Any previous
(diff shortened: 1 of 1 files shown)


dpipemazo (approved):
:+1: This is a huge win and will help in a lot of our multi-process repos.

PR #445, opened 2021-04-02
441/remove doc types

### LINKED ISSUE #441: Remove types from docs and add hints to variadic args
Some functions do not have type hints for their variadic args. We should add the type hints.

We also need to remove types from docstrings to maintain one source of truth now that type hints have been added to the majority of the codebase.

slaigai:
Closes #441

### DIFF
--- a/languages/python/atom/element.py
+++ b/languages/python/atom/element.py
@@ -104,8 +104,8 @@ class CommandHandlerMapBase(TypedDict):
 class CommandHandlerMap(CommandHandlerMapBase, total=False):
     """
     Using total=False so that `deserialize` is accepted as an optional key (but
-        not required) to check for deprecated legacy logic in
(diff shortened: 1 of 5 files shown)


dpipemazo (approved):
:+1: This really needed to get done. Thank you. 

Do you have an auto-formatter that works well for the docstrings/80 char comments? I had been doing it manually and was wondering if there was a better way.

slaigai:
>  This really needed to get done. Thank you.
> 
> Do you have an auto-formatter that works well for the docstrings/80 char comments? I had been doing it manually and was wondering if there was a better way.

I use the "Rewrap" extension on VSCode but TBH there was a lot of manual formatting I had to do on top of that. Rewrap takes care of the majority of the wrapping but it isn't perfect...

PR #453, opened 2021-04-27
Build Simplification + Upgrades

dpipemazo:
Now that we have CircleCI ARM executors + Docker build caching via `buildx` we can go ahead and remove the clunky, "base" portion of our atom builds. This portion was added to keep rebuild time/complexity low, but was decently confusing and cumbersome.

As such, the following has been done to the Dockerfile/CircleCI config:
1. Simplify down to one, multi-stage Dockerfile
2. Add 3x different "base" variants to the single, multi-stage Dockerfile
3. Add an arg `ATOM_BASE` which allows the user to specify which base variant they want their atom/nucleus built atop. This is enabled/possible due to a **super cool** optimization that's only enabled in the newer buildkit-based Docker frontends where it can parse out a Dockerfile and **not build unnecessary stages based on args/target**. If you try to run `docker build` on your local machine it will work but will build every possible stage, follow the instructions to [set up buildx](https://github.com/elementary-robotics/atom#arm-images) and use `docker buildx build`.
4. Update the build process to use the single Dockerfile and remove the "base" component
5. Slim down the atom variants to just `base`, `base-cv` and `base-cv-graphics`. `base-cv` brings in opencv + related CV processing libraries, `base-cv-graphics` brings in opencv, cuda and opengl/VNC capabilities.

In doing the above, support has also been added for ARM-based CUDA images. In doing this, the AMD64 CUDA base has been switched over to one of the NVIDIA [NGC containers](https://www.nvidia.com/en-us/gpu-cloud/containers/) as this is the main place where we can get ARM CUDA containers. For code consistency, we're planning to use the same NGC tag from both their AMD and ARM containers. As such, we need to upgrade to the latest and greatest, which includes:
- CUDA 11.3
- CUDNN 8
- Ubuntu 20.04
as there was no build with CUDNN 10 on Ubuntu 18.04 available at the moment.

In switching over the ARM CUDA containers to this newer build, the Python version auto-installed/shipped with Ubuntu 20.04 is 3.8.5. There was/is no super-easy way to downgrade this to 3.7.x on both AMD + ARM (deadsnakes PPA only works on AMD it seems), so the decision was made to try to upgrade the main Atom base container to Python 3.8.x as well. There seems to be no easy way to upgrade `debian:buster-slim` to 3.8, so the easiest/most consistent thing to do is to switch the Atom main base container from `debian:buster-slim` to `ubuntu:20.04`. This gets even greater similarity between the CUDA/non-CUDA builds and doesn't seem to add too much size/overhead.

As a result, of the above, we have the following major changes in Atom at the merge of this PR:
1. `metrics` now builds for ARM
2. All atom images are based off of Ubuntu 20.04
3. All atom images will ship with Python 3.8.5 (upgrade from 3.7.3)
4. CUDA containers will be build for AMD + ARM
5. CUDA containers will ship with CUDA 11.3 + CUDNN8 runtime
6. Build no longer requires "base" steps, all will rebuild when appropriate source is changed. 

We do [get some nice features in Python 3.8.5](https://docs.python.org/3/whatsnew/3.8.html), including the walrus operator and auto-magic f-string variable dereferencing.


.circleci/README.md
```diff
@@ -125,6 +125,39 @@ circleci orb publish promote elementaryrobotics/atom@dev:some-tag patch
 
 ### Release Notes
 
+#### [v0.3.0](https://circleci.com/orbs/registry/orb/elementaryrobotics/atom?version=0.2.0)
+
+##### New Features
+
+** MAJOR CHANGES **
```
slaigai, line 132: The spaces should be removed if the intent is to bold this: `**MAJOR CHANGES**`

Dockerfile
```diff
@@ -1,15 +1,115 @@
...
+#  rebuild this part of the file often and can rely on the cache
 #
 ################################################################################
 
-ARG BASE_IMAGE=elementaryrobotics/atom:base
-ARG PRODUCTION_IMAGE=debian:buster-slim
-FROM $BASE_IMAGE as atom-source
-
+#
+# ALL ARGS TO BE USED IN **ANY** FROM MUST OCCUR BEFORE THE **FIRST** FROM
```
slaigai, line 9: :+1:

docker-compose-dev.yml
```diff
@@ -9,7 +9,8 @@ services:
       dockerfile: Dockerfile
       target: nucleus
       args:
-        BASE_IMAGE: elementaryrobotics/atom:v2.7.1-base-stock-amd64
+        STOCK_IMAGE: ubuntu:focal-20210416
```
slaigai, line 12: `Dockerfile` uses `ubuntu:20.04`. Although the hashes indicate that this is the exact same image, I would be nice to have one tag used everywhere to be consistent.

slaigai:
Hmm, looks like `install_arrow.sh` is deprecated and could be removed. I didn't look too hard but I couldn't find a reference to this script in this repo and I see the logic duplicated in the `Dockerfile`.

README.md
```diff
@@ -193,35 +127,34 @@ git submodule update --init --recursive
...
 
 You can pass the following build arguments to the build as well. To do this,
 simply edit the `args` section of the `docker-compose-dev.yml` to include/update
 variables of your choice:
 
 | Arg | Default | Description |
 |-----|---------|-------------|
-| `BASE_IMAGE` | elementaryrobotics/atom:base | Which base atom image to build atop. See [Base Images](#base-images) for choices |
-| `PRODUCTION_IMAGE` | debian:buster-slim | Which image to use for the slimmed, production stage. Typically best to use the `BASE_IMAGE` that was used when building the `base` that you chose to use. |
+| `STOCK_IMAGE` | `ubuntu:focal-20210416` | Which image to build from. Should work with most things debian-based |
```
slaigai, line 144: It would be nice to consistently use either `ubuntu:20.04` or `ubuntu:focal-20210416` consistently in all places where `STOCK_IMAGE` is referenced.

Dockerfile
```diff
@@ -1,15 +1,115 @@
...
+           .. \
+  && make -j8 \
+  && make install
+ARG PYARROW_EXTRA_CMAKE_ARGS=""
+RUN cd /atom/third-party/apache-arrow/python \
+  && ARROW_HOME=/usr/local SETUPTOOLS_SCM_PRETEND_VERSION="0.17.0" python3 setup.py build_ext -j 8 --build-type=release --extra-cmake-args=${PYARROW_EXTRA_CMAKE_ARGS} install
+
 #
 # Redis itself
 #
```
slaigai, line 115: The comment below is stale. RTS is finally home in base :house:

```
#
# Redis time series module. Should eventually make its way
#   into the base, but for now can live in here
#
ADD ./third-party/RedisTimeSeries /atom/third-party/RedisTimeSeries
WORKDIR /atom/third-party/RedisTimeSeries
```

docker-compose-dev.yml
dpipemazo: Thanks for the call-out, that's a bug in the Dockerfile. `focal-20210416` is pegged, while `20.04` gets pushed each time there's a new release. We should be using pegged builds only. Will update/fix

README.md
dpipemazo: Should be `focal-20210416` in all places. Will update

Dockerfile
dpipemazo: Ah, that should wind up living further up in the Dockerfile then potentially. Goal is to have things which are heavy to build and change infrequently earlier on to improve caching performance.

docker-compose-dev.yml
slaigai: Oh right, I hadn't considered that `20.04` gets updated :+1:

ashule01 (commented):
Looks good from what I can tell! Thanks for all the documentation on it

slaigai (approved):
Upgrade path (with templates) was clearly documented and worked for me! Docs look good all around, great work :raised_hands: !

.circleci/README.md
```diff
@@ -125,6 +125,39 @@ circleci orb publish promote elementaryrobotics/atom@dev:some-tag patch
...
+
+- Updates atom orb with significant changes to leverage new ARM executors in CircleCI as the primary/only executor type. Cross-compiling ARM images on intel machines is no longer supported. This effectively deprecates the need for the `platform` argument that was previously prevalent in the build, as we will determine our build platform from the instance type we're running on, intel or ARM. As such, major rework was done to the orb to remove the `platform` parameter.
+- Brings in for the first time multi-arch deploys. The deploy stage will now be able to wait for both intel + arm builds to complete and then push both images under a single tag to Dockerhub. This means that you can use the same tags in intel + ARM, and docker will auto-pull the correct one for your platform.
+- Deprecates the need for the `variant` concept in the orb. Instead of pushing intermediate/build tags with `-<< variant>>-<< platform >>`, intermediate tags are now just `-$(arch)` where `$(arch)` is either `x86_64` or `aarch64` depending on the platform. This obfuscates the tagging a bit less and makes the intermediate tags more understandable on most repos which didn't need/use the variant concept.
```
aywaldron, line 136: Just to make sure I understand, intermediate/build images get platform tags but deployed images do not?

.circleci/atom.yml
```diff
@@ -439,123 +304,71 @@ commands:
...
+      stage:
+        type: string
+        default: ""
+      atom_base_arg:
+        type: string
+        default: ""
     steps:
+
+      # Enable the buildx builder
       - enable_buildx
```
aywaldron, line 364: This step is repeated on the same build machine for prod and test builds. Not harmful but probably unecessary

dpipemazo: It is. It made the code a bit cleaner/easier and I figured it wasn't harmful.

.circleci/README.md
dpipemazo: Yes, exactly. The way that we do multi-arch is with a manifest, so we push `build-X-x86_64` and `build-X-aarch64` during the build steps, then the multi-arch manifest is created during deploy to have `development-X` point to both of those images using `docker manifest create`

aywaldron: Ah ok got it, thanks for explaining

aywaldron: I think I'm missing it but where does the `arch` var get set?

dpipemazo: Linux provides this, you can try running it on your laptop.

aywaldron (approved):
This is great Dan, much cleaner and easier to use.

PR #466, opened 2021-05-11
Feature/improve orb build

dpipemazo:
Shows significant speedup (20m on large Docker images) by not loading into the host docker driver and then pushing but pushing directly from `buildx`.


aywaldron (approved):
Awesome :tada:

Issue #393, opened 2021-01-07
Move submodules to use HTTPS cloning instead of SSH


dpipemazo:
Non-authenticated github users cannot clone our submodules. We should clone all submodules with HTTPS instead of SSH

For example: https://github.com/elementary-robotics/atom/blob/latest/.gitmodules#L36

PR #468, opened 2021-05-19
467/jetson updates

dpipemazo:
- Upgrades Python to 3.8.10 on all builds
- Builds python from source to enable consistency in version
- Upgrades numpy to 1.20.3 to work better with newer Pytorch on ARM
- Pre-builds wheels for cupy, pytorch and torchvision on the Jetson Xavier NX platform using the l4t container to get us Python 3.8.10, Pytorch 1.8.1, cupy 8.6.0 and torchvision 0.9.1 in our Jetson docker containers.

After manual compile/install on the host we are seeing `torch.cuda.is_available()` returning `True` which is exciting.

Work left to do:
- [x] Update CUDA install to install from pre-built wheel on ARM and using pegged requirements.txt file on intel
- [x] test final CUDA container on Jetson Xavier NX host

This PR does start us in using `git lfs` on this repo to host the wheels. It should be configured so that by default the LFS files are not pulled. To explicitly pull the LFS files, you can run
```
git lfs pull --include="*" --exclude=""
```


dpipemazo:
This PR also closes #393 as it was needed to clone/build on the Jetson without logging in

PR #470, opened 2021-05-20
add __str__ and __repr__ to Response

slaigai:
Makes it much easier to print out a response

### DIFF
--- a/languages/python/atom/messages.py
+++ b/languages/python/atom/messages.py
@@ -100,16 +100,29 @@ class Response:
         self.err_str = err_str
 
     def __str__(self) -> str:
-        return f"""err_code: {self.err_code}
-err_str: {self.err_str}
-ser: {self.ser}
-data: {atom_ser.deserialize(self.data, method=self.ser)}"""
+        return (
+            f"err_code: {self.err_code}\nerr_str: {self.err_str}\nser: {self.ser}"
+            f"data: {self._get_data_str_or_type()}"
+        )
 
     def __repr__(self) -> str:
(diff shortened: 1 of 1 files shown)


languages/python/atom/messages.py
```diff
@@ -99,6 +99,18 @@ def __init__(
...
         self.err_str = err_str
 
+    def __str__(self) -> str:
+        return f"""err_code: {self.err_code}
+err_str: {self.err_str}
+ser: {self.ser}
+data: {atom_ser.deserialize(self.data, method=self.ser)}"""
+
+    def __repr__(self) -> str:
+        return f"atom.messages.Response(data={atom_ser.deserialize(self.data, method=self.ser).__repr__()}, " \
```
mvanveen, line 109: do we really want all of the deserialized data within __repr__ in the REPL?

languages/python/atom/messages.py
```diff
@@ -99,6 +99,18 @@ def __init__(
         self.err_code = err_code
         self.err_str = err_str
 
+    def __str__(self) -> str:
+        return f"""err_code: {self.err_code}
+err_str: {self.err_str}
```
dpipemazo, line 104: Pretty minor, but not a huge fan of how this looks. Would prefer the multiline f-string that we can see in the `repr()` function below. Haven't looked too much into how to do decent-looking multiline f-strings.

languages/python/atom/messages.py
dpipemazo: We might want to add an explicit case where if `self.ser` is `None` or `arrow` then we just print the type, otherwise yes if it's msgpack'd we likely want to see it deserialized here.

dpipemazo (approved):

PR #473, opened 2021-05-27
upgrade formatter python version to 3.8.4

### LINKED ISSUE #472: Upgrade Python in Formatter
Atom now ships with Python 3.8.10, it's time to upgrade the formatter to the latest and greatest (`flake8:3.8.4`).

slaigai:
Closes #472

### DIFF
--- a/utilities/formatting/Dockerfile
+++ b/utilities/formatting/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine/flake8:3.7.0
+FROM alpine/flake8:3.8.4
 
 ADD ./requirements.txt .
 


dpipemazo (approved):

slaigai:
@dpipemazo just noticed that we'll get a newly added error `F541 f-string is missing placeholders`. This will cause existing builds to fail build checks unless we ignore this by default. I googled around a bit and didn't readily find a reason not to ignore this. What are your thoughts?

dpipemazo:
@slaigai I think we should leave it enabled and fix the issues as they come up.

PR #477, opened 2021-06-07
Atom-cli: Fixed the atom-cli records command.

sgord512:
This commit fixes the atom-cli  records command for the cmdres subcommand and removes the records log subcommand.

### DIFF
--- a/languages/python/atom/element.py
+++ b/languages/python/atom/element.py
@@ -641,10 +641,7 @@ class Element:
             element_name: Name of the element to generate the id for.
             stream_name: Name of element_name's stream to generate the id for.
         """
-        if element_name is None:
-            return stream_name
-        else:
-            return f"stream:{element_name}:{stream_name}"
+        return f"stream:{element_name}:{stream_name}"
 
     def _make_metric_id(self, element_name: str, m_type: str, *m_subtypes: str) -> str:
         """
--- a/utilities/atom-cli/atom-cli.py
+++ b/utilities/atom-cli/atom-cli.py
@@ -26,7 +26,7 @@ class AtomCLI:
         )
         self.session = PromptSession(style=self.style)
         self.print_atom_os_logo()
-        self.serialization = "msgpack"
+        self.serialization : ser.SerializationMethod = "msgpack"
         self.cmd_map = {
             "help": self.cmd_help,
             "list": self.cmd_list,
@@ -60,7 +60,6 @@ class AtomCLI:
                 Can filter records from the last N seconds or from certain elements.
 
                 Usage:
-                  records log [<last_N_seconds>] [<element>...]
                   records cmdres [<last_N_seconds>] <element>..."""
             ),
             "cmd_command": cleandoc(
@@ -221,9 +220,7 @@ class AtomCLI:
             start_time = "0"
             elements = set(args[1:])
 
-        if mode == "log":
-            records = self.mode_log(start_time, elements)
-        elif mode == "cmdres":
+        if mode == "cmdres":
             if not elements:
                 print(usage)
                 print(
@@ -240,30 +237,9 @@ class AtomCLI:
             print("No records.")
             return
         for record in records:
-            print(self.format_record(record))
+            print(self.format_record(record))    
 
-    def mode_log(self, start_time, elements):
-        """
-        Reads the logs from Atom's log stream.
-
-        Args:
-            start_time (str): The time from which to start reading logs.
-            elements (list): The elements on which to filter the logs for.
-        """
-        records = []
-        all_records = self.element.entry_read_since(
-            None, "log", start_time, serialization=None
-        )
-        for record in all_records:
-            if not elements or record["element"].decode() in elements:
-                record = {
-                    key: (value if isinstance(value, str) else value.decode())
-                    for key, value in record.items()
-                }  # Decode strings only which are required to
-                records.append(record)
-        return records
-
-    def mode_cmdres(self, start_time, elements):
+    def mode_cmdres(self, start_time : str, elements : "set[str]"):
         """
         Reads the command and response records from the provided elements.
         Args:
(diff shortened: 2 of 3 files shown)


languages/python/atom/element.py
```diff
@@ -641,10 +641,7 @@ def _make_stream_id(self, element_name: str, stream_name: str) -> str:
             element_name: Name of the element to generate the id for.
             stream_name: Name of element_name's stream to generate the id for.
         """
-        if element_name is None:
-            return stream_name
-        else:
-            return f"stream:{element_name}:{stream_name}"
+        return f"stream:{element_name}:{stream_name}"
```
dpipemazo, line 644: I am not sure this PR should be changing this logic. Please be careful as this can affect others who use this function

dpipemazo (changes requested):

languages/python/atom/element.py
sgord512: The typechecker already requires element_name to be of type `str`, so I don't think any typed code can trigger this logic. The only caller in the atom codebase I'm aware of that could pass `None` to this function is `atom-cli`, and this change removes the calls to `entry_read_since` that result in the `_make_stream_id` call. (Hence the removal of this codepath.)

After this change, the only callers are in `element.py` which is typed, and `test_atom.py`. `test_atom.py` doesn't call the function with `None`, at least not directly.

Unless I'm unaware of a caller in the untyped part of the codebase, I think this is pretty safe. Do you think I should revert this change?

dpipemazo: That's a fair analysis; I hadn't considered the typechecker :+1: . 

I'm usually a bit hesitant to make this sort of change as the types were added recently and there may be code in some of our other microservices depending on this. I'm usually a bit hesitant to make small changes like this, but do appreciate that it could simplify the code.

dpipemazo: Taking a closer look, this is likely a great catch and a change we can make :+1:. After we double-check some of the internal codebases I think we can go ahead and merge this in. Thanks!

dpipemazo (approved):
:+1: Could you please put a comment up with the new/sample output of the `cmdres` command so that we can review? Otherwise, looking great :+1:

sgord512:
Here's an example of the output of `records cmdres atombot`: 
```shell
{
  "element": "atombot",
  "id": "1623031355107-0",
  "language": "Python",
  "type": "response",
  "version": "2.7.1"
}
{
  "element": "atombot",
  "id": "1623031355109-0",
  "language": "Python",
  "type": "command",
  "version": "2.7.1"
}
{
  "cmd": "version",
  "data": "",
  "element": "atombot",
  "id": "1623031355242-0",
  "type": "command"
}
{
  "cmd_id": "1623031355242-0",
  "element": "atombot",
  "id": "1623031355260-0",
  "timeout": "1000",
  "type": "response"
}
{
  "cmd": "version",
  "cmd_id": "1623031355242-0",
  "data": {
    "language": "Python",
    "version": 2.7
  },
  "element": "atombot",
  "err_code": 48,
  "err_str": "",
  "id": "1623031355264-0",
  "type": "response"
}
{
  "cmd": "healthcheck",
  "data": "",
  "element": "atombot",
  "id": "1623031355271-0",
  "type": "command"
}
{
  "cmd_id": "1623031355271-0",
  "element": "atombot",
  "id": "1623031355272-0",
  "timeout": "1000",
  "type": "response"
}
{
  "cmd": "healthcheck",
  "cmd_id": "1623031355271-0",
  "data": "",
  "element": "atombot",
  "err_code": "0",
  "err_str": "Everything is good",
  "id": "1623031355273-0",
  "type": "response"
}
{
  "element": "atombot",
  "id": "1623031383407-0",
  "language": "Python",
  "type": "command",
  "version": "2.7.1"
}
{
  "element": "atombot",
  "id": "1623031383407-0",
  "language": "Python",
  "type": "response",
  "version": "2.7.1"
}
{
  "cmd": "version",
  "data": "",
  "element": "atombot",
  "id": "1623031383460-0",
  "type": "command"
}
{
  "cmd_id": "1623031383460-0",
  "element": "atombot",
  "id": "1623031383475-0",
  "timeout": "1000",
  "type": "response"
}
{
  "cmd": "version",
  "cmd_id": "1623031383460-0",
  "data": {
    "language": "Python",
    "version": 2.7
  },
  "element": "atombot",
  "err_code": 48,
  "err_str": "",
  "id": "1623031383476-0",
  "type": "response"
}
{
  "cmd": "healthcheck",
  "data": "",
  "element": "atombot",
  "id": "1623031383485-0",
  "type": "command"
}
{
  "cmd_id": "1623031383485-0",
  "element": "atombot",
  "id": "1623031383486-0",
  "timeout": "1000",
  "type": "response"
}
{
  "cmd": "healthcheck",
  "cmd_id": "1623031383485-0",
  "data": "",
  "element": "atombot",
  "err_code": "0",
  "err_str": "Everything is good",
  "id": "1623031383487-0",
  "type": "response"
}
{
  "cmd": "version",
  "data": "",
  "element": "atombot",
  "id": "1623031427901-0",
  "type": "command"
}
{
  "cmd": "command_list",
  "data": "",
  "element": "atombot",
  "id": "1623031427910-0",
  "type": "command"
}
{
  "cmd": "version",
  "data": "",
  "element": "atombot",
  "id": "1623082168882-0",
  "type": "command"
}
{
  "cmd": "command_list",
  "data": "",
  "element": "atombot",
  "id": "1623082168916-0",
  "type": "command"
}
{
  "cmd": "move_left",
  "data": "",
  "element": "atombot",
  "id": "1623082177330-0",
  "type": "command"
}
```

utilities/atom-cli/atom-cli.py
```diff
@@ -242,59 +239,49 @@ def cmd_records(self, *args):
...
-            cur_records = self.element.entry_read_since(
-                None, stream, start_time, serialization=None
-            )
-            for record in cur_records:
-                for key, value in record.items():
+            streams[self.element._make_response_id(element)] = start_time
+            streams[self.element._make_command_id(element)] = start_time
+
+        # Read data
+        stream_entries = self.element._rclient.xread(streams)
```
kylebebak, line 256: Much better to have an `xread` call instead of changing the `entry_read_since` API so it can handle this use case. Nice work

languages/python/atom/element.py
kylebebak: Agreed that it's good to get rid of this and simplify this function as long as we're sure it's not being used by any atom client code. Not only do we simplify it, we don't encourage client code to use it in ways it likely shouldn't be used

UPDATE: just pulled latest for nearly all of our edge repos, then grepped to see where we're using `_make_stream_id`

```
➜  ElementaryRobotics rg _make_stream_id
asset-management/asset_management/download.py
655:            self.element._make_stream_id(

vision-processing/vision/handlers/stream.py
118:        stream_name = element._make_stream_id(

atom/languages/python/atom/element.py
516:        self._rclient.unlink(self._make_stream_id(self.name, stream))
636:    def _make_stream_id(self, element_name: str, stream_name: str) -> str:
911:        return self._redis_scan_keys(self._make_stream_id(element_name, "*"))
1972:            stream_id = self._make_stream_id(
2073:            stream_id = self._make_stream_id(element_name, stream_name)
2192:            stream_id = self._make_stream_id(element_name, stream_name)
2323:                self._make_stream_id(self.name, stream_name), vars(entry), maxlen=maxlen
2904:            stream_name = self._make_stream_id(element, stream)

atom/languages/python/tests/test_atom.py
46:            private_sn = element._make_stream_id(element.name, s)
199:        assert caller._make_stream_id("abc", "123") == "stream:abc:123"
```

It's used nowhere but atom, and it hasn't broken anything in atom, so this looks great =)

kylebebak (approved):

PR #483, opened 2021-06-19
Add ability to override element name on entry write

aywaldron:
Adds new parameter to `entry_write` which allows for overriding the stream's element name. Defaults to the calling element name if unspecified.

### DIFF
--- a/languages/python/atom/element.py
+++ b/languages/python/atom/element.py
@@ -501,20 +501,31 @@ class Element:
     def __repr__(self):
         return f"{self.__class__.__name__}({self.name})"
 
-    def clean_up_stream(self, stream: str) -> None:
+    def clean_up_stream(self, stream: str, element_name: str = None) -> None:
         """
         Deletes the specified stream.
 
         Args:
             stream: The stream to delete.
+            element_name: (optional) Name of element to make stream ID from.
+                If not specified, will use this element's name and check that
+                stream exists on this element.
         """
-        if stream not in self.streams:
-            raise RuntimeError(
-                "Stream '%s' is not present in Element "
-                "streams (element: %s)" % (stream, self.name),
-            )
-        self._rclient.unlink(self._make_stream_id(self.name, stream))
-        self.streams.remove(stream)
+        # Set default element name
+        if not element_name:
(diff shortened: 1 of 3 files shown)


dpipemazo (approved):
:+1: API looks great! Thanks for getting this in

languages/python/atom/element.py
```diff
@@ -501,20 +501,31 @@ def __init__(
     def __repr__(self):
         return f"{self.__class__.__name__}({self.name})"
 
-    def clean_up_stream(self, stream: str) -> None:
+    def clean_up_stream(self, stream: str, element_name: str = None) -> None:
```
kylebebak, line 504: 👍

languages/python/atom/element.py
```diff
@@ -2305,7 +2319,12 @@ def entry_write(
         # Get a metrics pipeline
         with MetricsPipeline(self) as pipeline:
 
-            self.streams.add(stream_name)
+            # Assign default element name if not specified
+            element_name = element_name if element_name else self.name
+
+            if element_name == self.name:
```
kylebebak, line 2325: Cool, no metrics if client code specifies its own element name

languages/python/atom/element.py
```diff
@@ -2305,7 +2319,12 @@ def entry_write(
         # Get a metrics pipeline
         with MetricsPipeline(self) as pipeline:
 
-            self.streams.add(stream_name)
+            # Assign default element name if not specified
+            element_name = element_name if element_name else self.name
```
kylebebak, line 2323: This can be written as `element_name = element_name or self.name`, but doesn't matter

kylebebak (approved):
Very nice =)

languages/python/atom/element.py
aywaldron: The metrics are always on the calling element, no matter what element name it specifies.

dpipemazo:
@aywaldron please let me know when you're ready to merge and we can go ahead and do so. This is great, thanks!

aywaldron:
@dpipemazo This should be ready -- I will merge. Thanks!

PR #493, opened 2021-07-13
Optionally use threads in command_loop

kylebebak:
Closes https://elementary.atlassian.net/browse/ERS-855

- adds `use_procs` arg to `command_loop`, which defaults to true, but can be set to false to use threads for command loop instead of processes
- removes the need to do this kind of stuff, e.g. in integration framework

```py
        command_thread = Thread(
            target=self.element._command_loop,
            args=(
                self.element._command_loop_shutdown,
                0,
            ),
            daemon=True,
        )
```

Dan and I agreed it would be better to go with the multi-element fix to https://elementary.atlassian.net/browse/ERS-851, especially because we want to hotfix this back into 2.3

Still, this is a good feature to add, and once it's verified working we can use this solve for wherever we want a multi-threaded command loop

### DIFF
--- a/languages/python/atom/element.py
+++ b/languages/python/atom/element.py
@@ -1138,11 +1138,11 @@ class Element:
 
     def command_loop(
         self,
-        n_procs: int = 1,
-        use_procs: bool = True,
+        n_workers: int = 1,
         block: bool = True,
         read_block_ms: int = 1000,
         join_timeout: Optional[float] = None,
+        use_procs: bool = True,
     ) -> None:
         """
         Main command execution event loop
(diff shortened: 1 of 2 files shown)


languages/python/atom/element.py
```diff
@@ -1167,33 +1171,18 @@ def command_loop(
...
-            self.logger.warning(
-                f"[element:{self.name}] Active thread count is currently {thread_count}.  Child command_loop "
-                "processes will only copy one active thread's state and therefore may not "
-                "work properly."
-            )
-
         self.processes = []
         for i in range(n_procs):
-            p = Process(
+            Cls = Process if use_procs else Thread
```
dpipemazo, line 1176: :+1: nice. This is a cool way of doing it.

languages/python/atom/element.py
```diff
@@ -1167,33 +1171,18 @@ def command_loop(
         if n_procs <= 0:
             raise ValueError("n_procs must be a positive integer")
 
-        # note: This warning is emitted in situations where the calling process
```
dpipemazo, line 1170: :+1: Thank you, this warning was a bit confusing and not meeting its intended purpose. I think we'll be OK without it.

dpipemazo (approved):

languages/python/atom/element.py
```diff
@@ -1138,6 +1139,7 @@ def wait_for_elements_healthy(
     def command_loop(
         self,
         n_procs: int = 1,
+        use_procs: bool = True,
```
aywaldron, line 1142: I think this API is slightly confusing now since we're using `n_procs` for both processes and threads, while `use_procs` indicates processes and not threads. Perhaps we could change the name of `n_procs` to be more generic and not include "procs"?

languages/python/atom/element.py
```diff
@@ -1154,6 +1156,8 @@ def command_loop(
             n_procs: Number of worker processes.  Each worker process will pull
                 work from the Element's shared command consumer group (defaults
```
aywaldron, line 1157: Should update docstring to indicate either processes or threads

aywaldron (commented):
This is great Kyle! Just a comment on the API naming which I think could be improved

languages/python/atom/element.py
dpipemazo: +1 on this, "workers" might be more generic.

kylebebak:
Changed name of `n_procs` arg to `n_workers`, and moved `use_procs` arg to end of `command_loop` signature to ensure client code can't confuse `block` and `use_procs` args

PR #506, opened 2021-09-30
Shared memory

iourielementary:
## Change description

This is initial pull request for shared memory lib

- Pegged pyright to 1.1.188 

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Security

### Code review 


.gitmodules
```diff
@@ -37,3 +37,7 @@
 [submodule "metrics/grafana-redis-datasource"]
 	path = metrics/grafana-redis-datasource
 	url = https://github.com/RedisTimeSeries/grafana-redis-datasource
+[submodule "utilities/shared_mem/lib/pybind11"]
```
dpipemazo, line 40: do we need pybind as a submodule if we're `wget`-ing it in the Dockerfile build?

Dockerfile
```diff
@@ -68,10 +68,31 @@ ENV PATH="/opt/venv/bin:$PATH"
...
+# Install Boost
+RUN cd /tmp && mkdir boost && cd boost && \
+    wget -q --no-check-certificate https://boostorg.jfrog.io/artifactory/main/release/1.77.0/source/boost_1_77_0.tar.gz && \
+    tar -xzf boost_1_77_0.tar.gz && \
+    cd boost_1_77_0 && \
+    ./bootstrap.sh --with-libraries=date_time && ./b2 && \
+    ./b2 install && cd ../..
+
+# Install PyBind
+RUN python3.8 -m pip install pytest && \
```
dpipemazo, line 81: `python3` should be used here. We want to make sure we're picking up the one in the `virtualenv` so it all gets installed in the right place.

Dockerfile
```diff
@@ -68,10 +68,31 @@ ENV PATH="/opt/venv/bin:$PATH"
...
+    cd /tmp && mkdir pybind && cd pybind && \
```
dpipemazo, line 82: this would be a bit more readable/maintainable for me if each logical "line" were on its own line. It would add a lot of lines to the file, but would be easier to diff/track changes.

dpipemazo (commented):
Looking great. Left a few comments on the first pass review here. A few comments that are likely on your radar before merge. 


Please also comment/document the C/C++ code. We follow the [Google Style Guide](https://google.github.io/styleguide/cppguide.html) most of the time, but do have some comments on C code [here](https://github.com/elementary-robotics/wiki/wiki/Coding-Style)

I'll take a closer look at the implementation soon, thanks for putting this up!

Dockerfile
aywaldron: Agreed we should use `python3`; also, do we need `pytest` in here or can that wait until the test stage of the image?

Dockerfile
```diff
@@ -103,6 +124,13 @@ RUN cd /atom/third-party/OpenBLAS \
   && make TARGET=${BLAS_TARGET_CPU} -j8 \
   && make PREFIX=/usr/local install
 
+#ARG BLAS_TARGET_CPU=""
+#RUN cd /tmp && mkdir openblas && cd openblas && \
+#    wget -q --no-check-certificate https://github.com/xianyi/OpenBLAS/releases/download/v0.3.17/OpenBLAS-0.3.17.tar.gz && \
+#    tar -xzf OpenBLAS-0.3.17.tar.gz && cd OpenBLAS-0.3.17 && \
+#    make TARGET=${BLAS_TARGET_CPU} -j8 &&\
+#    make PREFIX=/usr/local install
```
aywaldron, line 132: Can we remove this now?

aywaldron (commented):
This is great Iouri!
In addition to the example python scripts, could we add a README that documents the python API?

Dockerfile
dpipemazo: Not sure -- but I imagine if it were then `pip3 install pybind` would take care of it. We should likely remove this.

aywaldron: Agree -- either way, (`pip3 install pybind` or `pytest`), we should include it up at line 69 with the other pip installs

dpipemazo (commented):
@iourielementary could we please rename everything from `er_shmem` (i.e. Elementary Robotics shared memory) to `atom_shmem`? I think this is more in line with it being in the atom ASK and if/when we spin the shared memory into its own library it would be under the atom umbrella.

Thanks!

dpipemazo:
> Are there any objections to make return error codes negative? 

No, that's pretty standard in C I think.

utilities/shared_mem/lib/atom_shmem_meta.cpp
```diff
@@ -0,0 +1,334 @@
...
+
+    VALIDATE_GLOBAL_PTRS(ptr_tagged_deques,ptr_active_handles);
+
+    TaggedDeques::iterator it;
+    StrAllocator str_alloc(shmem_ptr->get_segment_manager());
+    CharString search_tag(tag,str_alloc);
+    if ((it=ptr_tagged_deques->find(search_tag)) != ptr_tagged_deques->end()) {
+        if ((int64_t)it->second.size() > max_tag_elmnts) {
+            ActiveHandles::iterator handle_it;
+            for (;(int64_t)it->second.size()>max_tag_elmnts;) {
```
slaigai, line 315: Is there a convention or rule of thumb for using `for(; condition ;)` instead of a `while` loop?

utilities/shared_mem/lib/atom_shmem.cpp
```diff
@@ -0,0 +1,339 @@
...
+            std::vector<int64_t> removed_handles;
+
+            ptr_upgrd_mutex->unlock_upgradable_and_lock();
+            if ((ret = atom_shmem_meta_trim_tag_queue(ptr_shmem,max_tag_elmnts,tag,removed_handles)) != META_OK) {
+                ptr_upgrd_mutex->unlock_and_lock_upgradable();
+                return ret;
+            }
+            ptr_upgrd_mutex->unlock_and_lock_upgradable();
+
+            for (std::vector<int64_t>::iterator it=removed_handles.begin();it!=removed_handles.end();it++) {
```
slaigai, line 176: I could be wrong but I think the range-based syntax can be used here since vectors support `begin` and `end` right? Not trying to say that it's necessary or anything, just checking my understanding!

```
for(std::vector<int64_t>::iterator it; removed_handles)
```

dpipemazo:
Exciting to see this close! @aywaldron and @iourielementary let's hold off on merging this until we have a few PRs get in:
- https://github.com/elementary-robotics/atom/pull/512
- The one where we enable CPU/disk metrics by default which will come toay or Monday.

Thanks for letting these get in before making this large change -- just want to reduce risk in shipping the value from those features.

Dockerfile
```diff
@@ -44,7 +45,9 @@ RUN apt-get update \
       flex \
       bison \
       curl \
-      pkg-config
+      pkg-config \
+ && ln -s /usr/bin/gfortran-7 /usr/bin/gfortran \
```
aywaldron, line 49: Lets make this a separate RUN command so we can still cache all the installs even if we change these links

Dockerfile
```diff
@@ -45,10 +46,10 @@ RUN apt-get update \
       flex \
       bison \
       curl \
-      pkg-config \
- && ln -s /usr/bin/gfortran-7 /usr/bin/gfortran \
- && ln -s `dirname \`find /usr/lib -iname libgfortran.so.4 -not -path /usr/lib/gcc\``/libgfortran.so.4 `dirname \`find /usr/lib -iname libgfortran.so.4 -not -path /usr/lib/gcc\``/libgfortran.so \
- && ln -s `dirname \`find /usr/lib -iname libgfortran.so.4 -not -path /usr/lib/gcc\``/libgfortran.so.4 /usr/local/lib/libgfortran.so.4
+      pkg-config
+
+RUN ln -s `dirname \`find /usr/lib -iname libgfortran.so.4 -not -path /usr/lib/gcc\``/libgfortran.so.4 `dirname \`find /usr/lib -iname libgfortran.so.4 -not -path /usr/lib/gcc\``/libgfortran.so \
```
aywaldron, line 51: Let's add a comment above this explaining why we are creating these links and copying the library to `/usr/local/lib`

.gitmodules
iourielementary: What's our policy? We are wget-ing both pybind and boost, but openblas and numpy are submoduled?

dpipemazo: Our policy is not to use `wget` as we can't trust the sources to remain there/stable/not-changing. We submodule everything.

Dockerfile
iourielementary: @aywaldron Anna, we need to install pytest here, as part of pybind build process executes unit tests using pytest.

utilities/shared_mem/lib/atom_shmem.cpp
iourielementary: I am using begin and end? Could you clarify your comment?

Dockerfile
iourielementary: Done

Dockerfile
iourielementary: Done

Dockerfile
iourielementary: Done

Dockerfile
aywaldron: Ah ok, and pybind itself won't install it? If so that's fine but let's still take Dan's suggestion and switch `python3.8` to `python3`

utilities/shared_mem/py_app/jpeg_test.py
```diff
@@ -0,0 +1,68 @@
+import sys
+
+import cv2
+import ershmem
```
aywaldron, line 4: Just reiterating Dan's comment above re: changing this name from `ershmem` to `atomshmem` or something similar

utilities/shared_mem/lib/atom_shmem_meta.cpp
iourielementary: Nope, this is just my way, as I never get while loop right first time

.gitmodules
iourielementary: ok, I'll work on submoduling then

utilities/shared_mem/py_app/jpeg_test.py
iourielementary: Done

Dockerfile
```diff
@@ -46,6 +48,9 @@ RUN apt-get update \
       curl \
       pkg-config
 
+RUN ln -s `dirname \`find /usr/lib -iname libgfortran.so.4 -not -path /usr/lib/gcc\``/libgfortran.so.4 `dirname \`find /usr/lib -iname libgfortran.so.4 -not -path /usr/lib/gcc\``/libgfortran.so &&\
```
aywaldron, line 51: Can we add a comment above this line explaining which libraries needs these links in order to compile, and that the `cp` into `usr/local/lib` is for the prod image?

utilities/shared_mem/lib/atom_shmem.cpp
slaigai: Just noting to myself that this could have been written differently but equivalently. Nothing to change here

Dockerfile
aywaldron: @dpipemazo, to follow up on this, Iouri and I worked on it for a bit and `pytest` is only used here to run pybind's unit tests during its build. Its not able to run using the venv python for some reason, but we also don't care because this is purely a C library and does not install any python packages that we need. All that matters is that the unit tests are run against the same python that we are using in atom, which they are. We would just need to update this call if we upgrade python in the future.

dpipemazo (commented):
Overall looks good -- did not go too much in-depth into the code itself as the unit tests are quite thorough.

It would be nice to see documentation on each function which describes:
1. General description
2. All arguments and their purpose

Thanks!

utilities/shared_mem/lib/README.md
```diff
@@ -0,0 +1,451 @@
...
+Sets of allocated memory chunks can be created by using tags. Client associates memory chunk with a particular set by provideing a tag during allocation call.
+Memory chunk created with a tag **must** be deleted with the same tag. If client does not want to associate allocated chunk with any set, the **null** should be 
+provided as tag value. Memory allocate with **null** tag must be freed with **null** tag as well. Behavior is undefined if this convention is not observed.
+
+### Synchronization
+Access to shared memory is synchronized by two mechanisms
+- Boost `managed_shared_memory` internal synchronization is used to synchronize access to available shared memory
+- Boots `named_upgradable_mutex` is used to synchronize access to sets meta data
+
+`named_upgradable_mutes` is used so read access can be opmtimized (multiple users), writes to data structures are exclusive.
```
kgelem, line 27: typo :)

kgelem (commented):
tests look great.

Dockerfile
```diff
@@ -72,6 +77,27 @@ RUN pip3 install wheel
...
 
+# Install Boost
+ADD ./third-party/boost /atom/third-party/boost
+RUN cd /atom/third-party/boost && \
+    ./bootstrap.sh --with-libraries=date_time && ./b2 && \
+    ./b2 install
+
+# Install PyBind
+ADD ./third-party/pybind11 /atom/third-party/pybind11
+RUN /usr/local/bin/python3.8 -m pip install pytest && \
```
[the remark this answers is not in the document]
aywaldron: Per the other comment thread above about this, `pytest` is just being used to run unit tests as part of `pybind`'s compilation process and whatever is calling it can't find packages in the venv, so if we just install `pytest` in the venv `pybind` will through an error and not finish building. Luckily in this case we don't really care that its running outside the venv because its not installing any python packages that we need to use later. `pybind` is purely a C library that just happens to run python unit tests as part of its build. We just want to make sure we are running those tests with the same python version that atom will use.

Dockerfile
```diff
@@ -72,6 +77,27 @@ RUN pip3 install wheel
 # C/C++ deps
 #
 
+# Install Boost
+ADD ./third-party/boost /atom/third-party/boost
```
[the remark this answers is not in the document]
aywaldron: The boost library is in a submodule in this repo at `/atom/third-party/boost`, so the version is whichever commit we have that submodule at. This way we guarantee and have fine-grained control over the version we're building.

Dockerfile
iourielementary: It is needed by LAPACK that as part of OpenBLAS. It couldn't find libgfortran.so at compile time, although libgfortran.so.4 was installed. Hence symbolic link.

Dockerfile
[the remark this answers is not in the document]
iourielementary: to Anna's point. Since we run on multiple architectures we don't want to be at the mercy of package management system versioning. What I mean is that we have no control and leverage to make sure that same version of the package is used across arm and x86. Compiling from sources resolves this right away.

utilities/shared_mem/lib/atom_shmem_meta.cpp
```diff
@@ -0,0 +1,403 @@
...
+static const std::string active_handles_name = "ActiveHandles";
+
+static std::unique_ptr<TaggedDeques, TaggedDequesDeleter> ptr_tagged_deques;
+static std::unique_ptr<ActiveHandles, ActiveHandlesDeleter> ptr_active_handles;
+
+//#define VERBOSE_EXCPT
+
+#ifdef VERBOSE_EXCPT
+#define PRINT_EXCEPTION(ex)                                                    \
+    std::cout << "Caught exception: " << ex.what()                             \
```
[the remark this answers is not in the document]
iourielementary: All C++ exception are re-thrown at the Python layer, and it application responsibility to log. In the library itself it was used for debugging only

utilities/shared_mem/lib/README.md
iourielementary: oh, thank yoy

utilities/shared_mem/c_app/test.c
```diff
@@ -0,0 +1,46 @@
+#include <stdint.h>
```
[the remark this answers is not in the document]
iourielementary: done

iourielementary:
> Overall looks good -- did not go too much in-depth into the code itself as the unit tests are quite thorough.
> 
> It would be nice to see documentation on each function which describes:
> 
> 1. General description
> 2. All arguments and their purpose
> 
> Thanks!

Done

aywaldron (approved):
:tada:  Congrats Iouri!

PR #508, opened 2021-11-02
redis-metrics: change to AOF save policy

dpipemazo:
- Data will be more durable to stack start/stops
- New git-ignored folder metrics/data which will persist
the AOF between launches
- Updates to docker-compose.yml to make development easier
- Added error-handling script on launch to asses AOF file, attempt to repair it, and if all else fails, delete it so that the metrics redis launches as we would expect. 

## Change description

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Security

### Code review 


aywaldron:
Thanks for getting this up @dpipemazo! Only call out is the [redis persistence docs](https://redis.io/topics/persistence) recommend an `fsync` policy of `appendfsync everysec` when using AOF as a compromise between speed and reliability -- right now its set to `appendfsync no` and could result in 30s of data loss.

dpipemazo:
> recommend an fsync policy of appendfsync everysec when using AOF as a compromise between speed and reliability -- right now its set to appendfsync no and could result in 30s of data loss.

Yeah, this is interesting and worth discussing. I think `appendfsync no` is the right strategy here. What happens when the fsync is not run is that the redis thinks it wrote to the file on disk, but in reality linux is buffering it in RAM and it hasn't been flushed to disk yet. If we start/stop our stack, we don't lose any of this data, since as far as userspace linux is concerned, the file is still intact. If we were to reboot or lose power, that's when we would lose data.

Since power cycles/reboots are fairly rare, and given we're running on SSDs in the field which have could TRIM issues/could wear down, I think it's better/safer from a performance perspective to leave the fsync off and let linux decide when we page to disk. We won't lose data unless we power cycle, and the data will persist as expected across docker-compose up/down operations.

aywaldron:
Thanks for explaining Dan -- I think this tradeoff makes sense for our use case.

PR #512, opened 2021-11-08
Ers 1692/server side defaults

dpipemazo:
## Change description

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Security

### Code review 


.circleci/docker-compose-circle.yml
```diff
@@ -15,6 +15,9 @@ services:
     environment:
       - "NUCLEUS_METRICS_MONITOR=FALSE"
       - "ATOM_USE_METRICS=TRUE"
+      - "ATOM_METRICS_DEFAULT_COMPACTION_POLICY="
```
dpipemazo, line 18: Remove these -- the defaults are in the Dockerfile.

docker-compose.yml
```diff
@@ -34,6 +34,9 @@ services:
       - "METRICS_MONITOR_PROCFS=/host/proc"
       - "ATOM_USE_METRICS=TRUE"
       - "NUCLEUS_METRICS_MONITOR=FALSE"
+      - "ATOM_METRICS_DEFAULT_COMPACTION_POLICY=avg:1M:1d;sum:1M:1d;max:1M:1d"
```
dpipemazo, line 37: Remove these -- defaults set in the Dockerfile

languages/python/atom/element.py
```diff
@@ -4109,16 +3963,39 @@ def metrics_create_custom(
...
-                key,
-                retention_msecs=retention,
-                labels=_labels,
-                duplicate_policy=duplicate_policy,
-                chunk_size=METRICS_DEFAULT_CHUNK_SIZE,
-            )
+
+            # If we have no rules, fall back to server-side defaults
+            if rules is None:
+                data = self._mclient.add(
```
dpipemazo, line 3974: Add comment here explaining why this is add and not create.

dpipemazo:
This PR does the following:

- Allows our RTS redis to set server-side defaults for retention length, compaction rules, and duplicate policy

- Updates atom + all built-in metrics to respect/use the server-side defaults, retention rules and duplicate policy

- Preserves the ability to override retention length (no breaking changes)

- Preserves the ability to override/set rules. Breaking changes here in the API -- moved from dictionary of `key : (agggregation, time_bucket, retention)` to list of `(agggregation, time_bucket, retention)`. Reason for this is to allow us to match the auto-created keys when using RTS defaults with our overrides to standardize key names for aggregated time series. RTS and atom auto-generated aggregated key name is `f"{key}_{AGG}_{time_bucket}"`.

- Switches default duplicate sample policy from LAST to SUM here. This is a subtle breaking change. Our metrics calls use auto-generated millisecond-level timestamps. So, if we make 2x independent calls on the same key in the same millisecond (which we do not have control over, by design), we will hit a "duplicate". Originally, RTS would throw an error here which was very suboptimal. We then switched the policy to "last", i.e. overwrite the original sample which was better as it didn't crash, but still loses data. We now are defaulting to `SUM`, which will sum all values with the same timestamp. `SUM` works well for counters and most metrics, however it can be quite misleading for timing metrics, where a policy of `LAST` or `MAX` may be better. We can/should discuss if this is the best default for Atom overall and we should be careful, when updating code with timing metrics, to choose a policy of either `LAST` or `MAX`. https://oss.redis.com/redistimeseries/configuration/#duplicate_policy

- Upgrades RTS to 1.4.13. This is the most recent stable build of 1.4. There's a newer version, 1.6, available, but it's still in "RC" stage so it's likely not the right call to upgrade now.

- Upgrades Redis to 6.2.6 + fixes `stream-node-max-*` misconfigurations which were deemed to be the root cause of streams not trimming themselves: https://github.com/redis/redis/issues/9469

config/nucleus/redis/redis-atom.conf
```diff
@@ -1550,8 +1550,8 @@ hll-sparse-max-bytes 3000
 # zero, the limit is ignored, so for instance it is possible to set just a
 # max entires limit by setting max-bytes to 0 and max-entries to the desired
 # value.
-stream-node-max-bytes 0
-stream-node-max-entries 0
+stream-node-max-bytes 4096
+stream-node-max-entries 100
```
aywaldron, line 1554: Could you explain the change from no limit to limits here and in the metrics conf below?

aywaldron (commented):
Thanks for the explanation here Dan. One high level question I have is can we use something like the `MetricsLevel`s to categorize metrics as types that have their own default retention/aggregation/duplication policies. E.g. we'll almost always want timing metrics to aggregate using MAX and duplicate with MAX, and we'll almost always want count metrics to aggregate using SUM and duplicate using SUM. If we could have the concept of these "types" in atom I think it would cut down on a lot of work changing over all the application code and reduce our chances to make mistakes setting these policies by hand.

dpipemazo:
> can we use something like the MetricsLevels to categorize metrics as types that have their own default retention/aggregation/duplication policies. E.g. we'll almost always want timing metrics to aggregate using MAX and duplicate with MAX, and we'll almost always want count metrics to aggregate using SUM and duplicate using SUM. If we could have the concept of these "types" in atom I think it would cut down on a lot of work changing over all the application code and reduce our chances to make mistakes setting these policies by hand on each metric.

This is a great idea! If it's OK let's add this as a follow-up; I am not sure I have the time to implement this for now. We'll still be able to override on a per-metric basis for the ones we're creating in the elements for now.

dpipemazo:
@aywaldron if we care most about timing metrics for now we can also make the default policy by MAX or LAST; perhaps you can let me know what you think is best.

aywaldron:
@dpipemazo yeah I think `MAX` would be a good default duplicate policy for now.

dpipemazo:
>  think MAX would be a good default duplicate policy for now. I'll make tickets for the follow on work.

:+1: sounds good, will do

config/nucleus/redis/redis-atom.conf
```diff
@@ -343,7 +341,7 @@ rdb-del-sync-files no
 # The Append Only File will also be created inside this directory.
 #
 # Note that you must specify a directory here, not a file name.
-dir ./
+dir /shared
```
aywaldron, line 344: Will this require an ansible change to where the edge metrics volume is mounted?

aywaldron (approved):
Looks great Dan, thanks for getting this all working

config/nucleus/redis/redis-atom.conf
dpipemazo: no, config change should have no discernible effect. It's just cleanup so that if people do turn on persistence/backups on the nucleus (which are/should be disabled right now), it'll go into a reasonable folder.

PR #527, opened 2022-02-15
Jetson Performance Optimizations

dpipemazo:
## Change description

Improve perfomance on Jetson targets through realtime scheduling + fire-and-forget metrics operations.  

Detail:
- Elements now can use a "fire and forget" mode when interacting with metrics and/or redis streams. Default is to do this with metrics but not with streams. Streams fire-and-forget can be enabled on a case-by-case basis in the `entry_write` call. All fire-and-forget can be turned off by setting the `ATOM_USE_FF` environment variable to `FALSE`. 
- Adds a script, `/usr/local/bin/update_scheduling.sh` which will use `chrt` or `nice` to change the scheduling algorithm/priority/niceness based off of environment variables. This will be unused by default, but if an element adds it into their launch path, it will cause the process launched and all children to inherit the environment variables for scheduling.
- Upgrades grafana-redis-datasource to work with newer Grafana. These changes broke stages in the build which were cached so we had to fix this now
- Switches to using a version of `redis-py` which has a "fire and forget" mode. Installs `redis-py` from a fork that we now host (https://github.com/elementary-robotics/redis-py) which is at 4.1.0 + our modifications
- `redis-py` 4.1.0 deprecated the `RedisTimeSeries` python library and broke backwards compatibility. Upgraded all code to use the new `redis-py` version of RTS commands and remove the `RedisTimeSeries` library.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Security

### Code review 


languages/python/atom/element.py
```diff
@@ -126,6 +124,9 @@ class CommandHandlerMap(CommandHandlerMapBase, total=False):
 ATOM_LOG_DIR = os.getenv("ATOM_LOG_DIR", "")
 ATOM_LOG_FILE_SIZE = int(os.getenv("ATOM_LOG_FILE_SIZE", LOG_DEFAULT_FILE_SIZE))
 
+# Control whether or not we use fire and forget
+ATOM_USE_FF = os.getenv("ATOM_USE_FF", "TRUE") == "TRUE"
```
slaigai, line 128: NIT: We've spent some time tracing configuration errors back to these bool env vars. Would be nice to check against `.upper()` to avoid case errors. For example, `True` would result in a `ATOM_USE_FF = False`

languages/python/atom/element.py
```diff
@@ -243,10 +244,13 @@ def __init__(
             self._redis_connection_timeout > 0
         ), "timeout must be positive and non-zero"
         self.streams: set[str] = set()
-        self._rclient: redis.StrictRedis
+        self._rclient: redis.Redis
+        self._rclient_ff: redis.Redis
         self._command_loop_shutdown = multiprocessing.Event()
-        self._rpipeline_pool: "Queue[Pipeline]" = Queue()
-        self._mpipeline_pool: "LifoQueue[RedisTimeSeriesPipeline]" = LifoQueue()
+        self._rpipeline_pool: "LifoQueue[Pipeline]" = LifoQueue()
```
slaigai, line 250: For my understanding, why was the `Queue` swapped out for a `LifoQueue`? I'm not sure whether the order of pipelines in the pool matters at all so it seems like an OK change. Just making sure we didnt miss something.

slaigai (approved):
A nit and a question. Otherwise looks good to me!

languages/python/atom/element.py
dpipemazo: good call-out! Will do this.

Dockerfile
```diff
@@ -376,10 +375,16 @@ RUN cd /atom/languages/cpp \
 # Python client
 #
 
-# Build and install the python library
-# Add and install requirements first to use DLC
+# Install redis-py
+ADD ./languages/python/third-party/redis-py /atom/languages/python/third-party/redis-py
+WORKDIR /atom/languages/python/third-party/redis-py
+RUN python3 setup.py install
```
aywaldron, line 381: Is the installation via pip insufficient?

aywaldron (approved):
Looks great, thanks Dan

ashule01 (approved):
Looks awesome, thanks Dan!

Dockerfile
dpipemazo: yes, we now have it as a [submodule](https://github.com/elementary-robotics/redis-py) since we're running a fork which implements the fire and forget mode. 

I'm not sure if I'll try to upstream/merge the "fire and forget" mode into `redis-py` yet so we might be running this fork for a while.

languages/python/atom/metrics.py
```diff
@@ -11,7 +11,6 @@
 )
 from atom.element import Element
 from redis.client import Pipeline
-from redistimeseries.client import Pipeline as RedisTimeSeriesPipeline
```
aywaldron, line 14: bump

languages/python/tests/test_atom.py
```diff
@@ -89,9 +88,9 @@ def _element_cleanup(self, element):
 
     def _get_redis_client(self):
         if TEST_REDIS_HOST is not None:
-            client = redis.StrictRedis(host=TEST_REDIS_HOST, port=TEST_REDIS_PORT)
+            client = redis.Redis(host=TEST_REDIS_HOST, port=TEST_REDIS_PORT)
```
aywaldron, line 91: bump

PR #541, opened 2022-04-05
Adding API functionality for shared memory

ashule01:
...reference_create, reference_get, and reference_delete for shared memory

## Change description

> Adding functionality for creating, getting, and deleting shared memory references through the current reference API. These changes are backwards compatible by creating redis references if no ref_type is specified

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Security

### Code review 


iourielementary (approved):

languages/python/atom/element.py
```diff
@@ -2819,7 +2868,186 @@ def reference_create(
...
+
+        Returns:
+            List of references corresponding to the arguments passed
+        """
+
+        ref_ids = []
+        # Get a metrics pipeline
+        with MetricsPipeline(self) as pipeline:
+
+            # _pipe = self._rpipeline_pool.get()
```
jgauth, line 2902: delete?

languages/python/atom/element.py
```diff
@@ -567,6 +587,49 @@ def clean_up_stream(self, stream: str, element_name: str = None) -> None:
...
+        except atomshmem.AtomShMemException as e:
+            self.logger.info(f"Atom shared memory creation exception: {e}")
+        except Exception as e:
+            raise ValueError(e)
+
+    def _shared_mem_destroy(self, name: Optional[str] = None) -> int:
+        """Destroys allocated shared mem"""
+
+        if name is None:
+            name = ATOM_SH_MEM_NAME
```
jgauth, line 606: It seems like this could be replaced with setting the name argument to default to `ATOM_SH_MEM_NAME` instead of `None`

languages/python/atom/element.py
```diff
@@ -2819,7 +2868,186 @@ def reference_create(
...
+                )
+                ref_ids.append(key)
+
+            self.metrics_timing_end(
+                self._reference_create_metrics["serialize"], pipeline=pipeline
+            )
+
+            # Write data
+            self.metrics_timing_start(self._reference_create_metrics["data"])
+            self.metrics_timing_end(
```
jgauth, line 2941: nothing happens between timing start and end?

aywaldron (changes requested):
Overall functionality looks great; suggestions are mainly around error handling, documentation, testing and style. Please make sure we have types defined in all function parameters.

One code change that I think could be helpful is to have classes for the different reference types, that define their unique reference create/get/delete functions. We do something like this with the serialization options [here](https://github.com/elementary-robotics/atom/blob/latest/languages/python/atom/serialization.py#L15). For example:
```
class AtomReference:
    @classmethod
    def reference_create(cls, data):
        raise(NotImplementedError)

    @classmethod
    def reference_get(cls, key):
        raise(NotImplementedError)
        
    @classmethod
    def reference_delete(cls, key):
             raise(NotImplementedError)

class RedisReference(AtomReference):
     # override the above methods with the methods you've defined on Element
     
class SharedMemReference(AtomReference):
     # override the above methods with the methods you've defined on Element
```

dpipemazo (changes requested):

languages/python/atom/element.py
```diff
@@ -3025,7 +3297,107 @@ def reference_get(
...
+                    handle = int(key_split[key_split.index("reference") + 2])
+                    try:
+                        data = atomshmem.atom_shmem_get(handle)
+                    except Exception as e:
+                        self.logger.warning(
+                            f"An exception occurred when getting data for handle {handle}: {e}"
+                        )
+
+                    if type(data) is not np.ndarray:
+                        raise ValueError(f"Invalid response from shared_mem: {data}")
```
slaigai, line 3339: NIT: Error string should indicate that data must be `np.ndarray` when using shared mem. Though the string is accurate, it does not explain _why_ the response is invalid so that the reader can understand/fix the issue.

languages/python/atom/element.py
```diff
@@ -3092,10 +3488,24 @@ def reference_delete(self, *keys) -> tuple[bool, list[str]]:
...
+
         with MetricsPipeline(self) as metrics_pipeline:
             self.metrics_timing_start(self._reference_delete_metrics["delete"])
 
-            success, failed = self._redis_key_delete(*keys)
+            success = True
+            failed = []
+            if redis_keys:
+                redis_success, redis_failed = self._redis_key_delete(*redis_keys)
+                success = success and redis_success
```
slaigai, line 3500: :+1:

languages/python/atom/element.py
ashule01: Good call out!

languages/python/atom/element.py
```diff
@@ -567,6 +587,49 @@ def clean_up_stream(self, stream: str, element_name: str = None) -> None:
...
+
+    def _shared_mem_cleanup_tag(self, tag: str) -> int:
+        """Deallocates data for all handles in a tag"""
+
+        ret = atomshmem.atom_shmem_tag_cleanup(tag)
+
+        return ret
+
+    def _shared_mem_cleanup_all_tags(self) -> tuple[bool, dict[str, int]]:
+        """Cleans up all tags associated with this element"""
```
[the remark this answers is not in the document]
aywaldron: I think we should take Dan's suggestion below and have tags scoped to the element that creates them so we don't have multiple elements using the same tag. If we do this, we should also add the element name to the tag name the user specifies before passing it onto the shared mem library, e.g. if the basler element wants to create a reference under the `image` tag, we construct the tag `basler:image` to pass to the shared mem lib.

languages/python/atom/element.py
```diff
@@ -769,8 +834,9 @@ def _make_reference_id(self, key: Optional[str] = None) -> str:
         Returns:
             Full reference key that includes element name
         """
+        tag = tag if tag else self.name
```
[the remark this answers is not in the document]
aywaldron: I think this is a good idea and allows us to cleanup tags on element cleanup.

languages/python/atom/element.py
```diff
@@ -2819,7 +2868,186 @@ def reference_create(
...
+                # logic to determine what the max_len should be
+                # always take the longest input as max_len
+                if tag_name in self.shared_mem_tag_max_len:
+                    tag_max_len = max(max_len, self.shared_mem_tag_max_len[tag_name])
+                else:
+                    tag_max_len = max_len
+
+                try:
+                    alloc_id = atomshmem.atom_shmem_alloc_init(
+                        tag_name, tag_max_len, datum, datum.size
```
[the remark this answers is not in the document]
aywaldron: Great catch -- is there another more generic way to get data size we could use here than calling `.size`? Our first use case for this API is numpy arrays but I'm not sure we want/need to make it specific to that data type.

languages/python/atom/element.py
```diff
@@ -3025,7 +3297,107 @@ def reference_get(
...
+        # to avoid having to reorder it based on the order
+        # of the input keys
+        return_data = [None] * len(keys)
+        if shared_mem_keys:
+            return_data = self._reference_get_shared_mem(
+                shared_mem_keys, return_data=return_data
+            )
+
+        if redis_keys:
+            return_data = self._reference_get_redis(
```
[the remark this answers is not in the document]
ashule01: The idea would be that you can retrieve multiple keys regardless of their type in a single call. Because return data is passed in to both redis and shared_mem reference_get functionality and the return data is stored in the exact same order that the keys are passed in (via the index of the key in the original keys array), you wouldn't end up overwriting it. I ended up implementing it this way because I wanted to avoid having to copy data over from different return data lists from the different reference types to be able to return a single value

languages/python/atom/element.py
```diff
@@ -2940,9 +3168,55 @@ def _reference_get_init_metrics(self) -> None:
...
+        Return:
+            key type: AtomRefType enum of either shared mem or redis
+        """
+
+        key_split = key.split(":") if type(key) == str else key.decode().split(":")
+
+        if "type" in key_split:
+            # specified type could be either redis or shared_mem
+            ref_type_val = key_split[key_split.index("type") + 1]
+            return AtomRefType(ref_type_val)
```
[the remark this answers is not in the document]
ashule01: So you need (...) here instead because ref_type_val is the value assigned to the enumeration (ex: "shared_mem") instead of the name of the enum (ex: SHARED_MEM), so by accessing it via parentheses, you are getting the name of that AtomRefType for the given val

languages/python/atom/element.py
```diff
@@ -3016,7 +3287,8 @@ def reference_get(
                 )
 
                 # Deserialize the data
-                deserialized_data.append(
+                return_data_index = int(keys[key])
```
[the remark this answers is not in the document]
ashule01: Good call out on this -- so in order to avoid getting the data from shared_mem and redis keys separately then having to recombine the data into a single list to return from reference_get, keys in this function is a dictionary made up of the keys and their indices in the original keys array (ex: {"key":index_of_key_in_original_keys_array}) which allows us to return the data in the proper order without having to copy to a new array when reordering it. The naming is unclear though and not having typehints definitely makes this confusing as well. I'll add both

languages/python/atom/element.py
```diff
@@ -3025,7 +3297,107 @@ def reference_get(
...
+
+        Returns:
+            List of items corresponding to each reference key passed as an
+            argument
+        """
+
+        # Initialize metrics
+        self._reference_get_init_metrics()
+
+        shared_mem_keys, redis_keys = self._split_keys_by_ref_type(*keys)
```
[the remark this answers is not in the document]
ashule01: So we could combine for sure but the small hiccup is that for redis keys, the key is "gotten" but not until we run _pipe.execute() does it actually get all the data. So the pipeline for getting then properly formatting the data between shared_mem and redis are slightly different

- shared_mem: get data for each key individually and add it to the return data as we get it
- redis: get the data for all keys at the same time then properly format/deserialize it by looping through again then adding it to the return data

So to answer, we could but it does make combining the two kind of awkward since we don't want to fetch the redis results until we've added them all to the get queue via _pipe.get(key)

languages/python/atom/element.py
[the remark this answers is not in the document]
slaigai: Ohhhhhh you're right. I learned something today :)

languages/python/atom/element.py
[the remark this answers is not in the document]
slaigai: Oh gotcha, so the `return_data` going into the function is the same object reference that gets returned. Makes sense now that you point it out! I didn't read carefully enough to pick that up

languages/python/atom/element.py
aywaldron: @ashule01 Is the shared mem lib doing a check for `np.ndarray` type? Or does it allow other data types as well? I can't remember what we landed on there.

languages/python/atom/element.py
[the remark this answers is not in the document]
ashule01: So this is specified in the shared_mem library that it has to be an numpy array made up of ints -- definitely means I should add some more data type checking earlier in the process, but we are currently restricted to only numpy arrays of ints for shared memory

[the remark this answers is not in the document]
aywaldron: Ah I see, thanks for clarifying. Can disregard by other comment asking the same question :)

languages/python/atom/element.py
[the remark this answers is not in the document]
aywaldron: Ok, that makes sense. Let's stick with your current implementation then but maybe try to make the return data usage a bit more clear to a reader since it seems there's some confusion over how exactly its working.

languages/python/atom/element.py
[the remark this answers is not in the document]
aywaldron: Yeah, including more information in the docstring may also help since right now it suggests `keys` is just a list of keys.

languages/python/atom/element.py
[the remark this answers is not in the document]
ashule01: I'll definitely add a comment describing this though because it is definitely not obvious how this is working! Thanks for the comment on this :)

languages/python/atom/element.py
```diff
@@ -3025,7 +3363,105 @@ def reference_get(
...
+
+            # Get the data
+            self.metrics_timing_start(self._reference_get_metrics["data"])
+            for key in keys:
+                key_split = key.split(":")
+                data = None
+                if "reference" in key_split:
+                    handle = int(key_split[key_split.index("reference") + 2])
+                    try:
+                        data = atomshmem.atom_shmem_get(handle)
```
dpipemazo, line 3399:
```
data, w, h, d = atom_shmem_get(...)
data.dtype = numpy.uint8
data.shape = (w, h, d)
data.strides = (h*3, h, 1)
```

languages/python/atom/element.py
```diff
@@ -2819,7 +2939,170 @@ def reference_create(
...
+                # logic to determine what the max_len should be
+                # always take the longest input as max_len
+                if tag_name in self.shared_mem_tag_max_len:
+                    tag_max_len = max(max_len, self.shared_mem_tag_max_len[tag_name])
+                else:
+                    tag_max_len = max_len
+
+                try:
+                    alloc_id = atomshmem.atom_shmem_alloc_init(
+                        tag_name, tag_max_len, datum, datum.size
```
dpipemazo, line 2990:
```
alloc_id = atomshmem.atom_shmem_alloc_init(
    tag_name, tag_max_len, datum, datum.size, w, h, d)
```

where:
```
w, h, d = np.shape()
```

languages/python/atom/element.py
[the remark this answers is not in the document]
ashule01: I implemented this by setting it up as {element name}-{additional tag} with additional tag defaulting to "default" so you'd get something like `basler-default` if no additional tag is passed.

languages/python/atom/element.py
```diff
@@ -759,18 +873,39 @@ def _metrics_validate_labels(self, labels: dict) -> None:
...
         key = key if key else str(uuid.uuid4())
-        return f"reference:{self.name}:{key}"
+
+        ref_key = f"reference:{tag}:{key}"
+
+        # We are now going to add the ref_type to all new keys
+        ref_key += f":type:{ref_type.value}"
+
+        if serialization:
+            ref_key += f":ser:{serialization}"
```
aywaldron, line 906: If we are using serialization of "none" (I believe the default for most of the API) does it still get included here? We would want it to.

languages/python/atom/element.py
```diff
@@ -759,18 +873,39 @@ def _metrics_validate_labels(self, labels: dict) -> None:
...
 
         Returns:
             Full reference key that includes element name
         """
+        tag = tag if tag else self.name
         key = key if key else str(uuid.uuid4())
-        return f"reference:{self.name}:{key}"
+
+        ref_key = f"reference:{tag}:{key}"
+
```
aywaldron, line 901: My understanding was we wanted to do `reference:{self.name}:{tag}:{key}` for shared mem references; I don't think this allows that?

languages/python/atom/element.py
```diff
@@ -2819,7 +2952,214 @@ def reference_create(
...
+                if len(datum_shape) > 3:
+                    e = "Failed to alloc: Datum cannot be greater than a \
+                        3 dimensional array"
+                    failed_allocs.append(e)
+                    continue
+
+                if not all(isinstance(x, np.uint8) for x in datum.flatten()):
+                    e = "Failed to alloc: Each value in datum must be of \
+                        type uint8"
+                    failed_allocs.append(e)
```
aywaldron, line 3009: :+1:

languages/python/atom/element.py
```diff
@@ -2819,7 +2952,214 @@ def reference_create(
...
+                    continue
+
+                # adding the element name to the tag to avoid any tag
+                # overlap between elements
+                tag_name = self.name + "-"
+                tag_name += tag if tag is not None else "default"
```
aywaldron, line 3015: Oh I see the element name is being added here. Makes sense!

languages/python/atom/element.py
```diff
@@ -2819,7 +2952,214 @@ def reference_create(
...
+                    tag_max_len = max_len
+
+                try:
+                    # default w/h/d to 0 if datum.shape < 3
+                    shape = [0, 0, 0]
+                    for i, val in enumerate(datum.shape):
+                        shape[i] = val
+                    print(
+                        f"THIS IS THE SHAPE: {shape} of data {datum} with size {datum.size}"
+                    )
```
aywaldron, line 3030: Perhaps this should be a DEBUG log instead of a print statement?

languages/python/atom/element.py
```diff
@@ -2819,7 +2952,214 @@ def reference_create(
...
+                if tag_name in self.shared_mem_tag_max_len:
+                    tag_max_len = max(max_len, self.shared_mem_tag_max_len[tag_name])
+                else:
+                    tag_max_len = max_len
+
+                try:
+                    # default w/h/d to 0 if datum.shape < 3
+                    shape = [0, 0, 0]
+                    for i, val in enumerate(datum.shape):
+                        shape[i] = val
```
aywaldron, line 3027: :+1:

languages/python/atom/element.py
```diff
@@ -3025,7 +3425,118 @@ def reference_get(
...
+            List of items corresponding to each reference key passed as an
+            argument
+        """
+
+        # Initialize metrics
+        self._reference_get_init_metrics()
+
+        shared_mem_keys, redis_keys, unidentified_keys = self._split_keys_by_ref_type(
+            *keys
+        )
```
aywaldron, line 3519: :+1:

languages/python/atom/element.py
```diff
@@ -2940,9 +3280,70 @@ def _reference_get_init_metrics(self) -> None:
...
+        """
+
+        shared_mem_keys = {}
+        redis_keys = {}
+        unidentified_keys = {}
+        for i, key in enumerate(keys):
+            key = key if type(key) == str else key.decode()
+            ref_type = self._get_reference_type(key)
+            if ref_type is None:
+                unidentified_keys[key] = i
```
aywaldron, line 3335: We should probably never hit this because we assume redis type if "type" isn't in the key, right?

languages/python/tests/test_atom.py
```diff
@@ -3107,6 +3212,12 @@ def test_set_pop_n_max(self, caller):
             passed = True
         assert passed == True
 
+    def test_destroy_shared_mem(self, caller):
+        caller, caller_name = caller
+        ret = caller._shared_mem_destroy()
+
+        assert ret == True
```
aywaldron, line 3219: Should we test that we can't add a piece of data to it now that its deleted?

utilities/scripts/clear_shared_mem.py
```diff
@@ -0,0 +1,33 @@
...
+    # to destroy shared mem, we have to have the pointer to it
+    # which we get by trying to create/open it even if it
+    # already exists
+    try:
+        atomshmem.atom_shmem_create_open(ATOM_SH_MEM_NAME, 4096)
+    except Exception as e:
+        if "ATOM_SHMEM_EXISTS" in str(e):
+            pass
+        else:
+            print(f"Error in getting shared mem pointer: {str(e)}")
```
aywaldron, line 19: Should we indicate that we can't delete it here? To make it clear that shared memory was not wiped on startup.

utilities/scripts/clear_shared_mem.py
```diff
@@ -0,0 +1,33 @@
...
+            return
+
+    # now we will destroy the shared memory
+    try:
+        atomshmem.atom_shmem_destroy(ATOM_SH_MEM_NAME)
+    except Exception as e:
+        if "ATOM_SHMEM_DOES_NOT_EXIST" in str(e):
+            return
```
aywaldron, line 27: Let's print that we were successful in clearing it out

aywaldron (changes requested):
Looks great Allison, thanks for making all the changes! Couple minor suggestions and questions left, overall looks very solid and test coverage looks very good :tada:

languages/python/atom/element.py
aywaldron: NVM this comment, I saw how its working below.

ashule01: Yeah definitely open to talking about the "-" vs ":" syntax because colon would definitely cleaner but would make the parsing a little clunkier since we want to include the name and the tag in the shared memory tag -- open to doing it either way

languages/python/atom/element.py
ashule01: oh shoot apologies, that was from testing  -- I'll clean it up and make it debug

Dockerfile
```diff
@@ -518,6 +521,9 @@ RUN mkdir /metrics
 ADD config/nucleus/launch.sh launch.sh
 ADD config/nucleus/launch_metrics_monitor.sh launch_metrics_monitor.sh
 
+# Add in script to clear shared memory
+ADD ./utilities/scripts/clear_shared_mem.py /usr/local/bin/clear_shared_mem.py
```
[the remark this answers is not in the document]
ashule01: Good call, and yeah it doesn't need any further requirements

languages/python/atom/element.py
ashule01: "ser:none" will be included for any reference created in reference_create:
For redis, serialization gets specified here: https://github.com/elementary-robotics/atom/blob/281b8403c698aa18d4330b7de2651ced8fd6eb4a/languages/python/atom/element.py#L2929
And for shared mem, serialization is always none: https://github.com/elementary-robotics/atom/blob/281b8403c698aa18d4330b7de2651ced8fd6eb4a/languages/python/atom/element.py#L3050

But in reference_create_from_stream, the serialization was not previously specified in the key: https://github.com/elementary-robotics/atom/blob/281b8403c698aa18d4330b7de2651ced8fd6eb4a/languages/python/atom/element.py#L3236

aywaldron: Ok that all sounds fine, thanks for checking!

languages/python/atom/element.py
aywaldron: I think what you have is good! so it'll look like `reference:camera-image:1234`

ashule01: Yup!

languages/python/tests/test_atom.py
```diff
@@ -3218,6 +3218,14 @@ def test_destroy_shared_mem(self, caller):
```
aywaldron, line 3218: Does this test destroy the shared memory for all subsequent tests? Might want to explicitly document that if so.

aywaldron (approved):
Looks great Allison, thanks for working through all the feedback. One suggestion on documentation in the unit tests, otherwise approving.

aywaldron (changes requested):
Actually one more request to update the atom docs in [here](https://github.com/elementary-robotics/atom/tree/latest/doc/source/includes) with the API changes.

languages/python/tests/test_atom.py
ashule01: Updated the test accordingly because the shared memory attempts to get created at the start of every test because of the creation of the "test" element. Thanks for the call out here

dpipemazo (approved):
My comments are mainly nits -- looks great! Excited to try this out.

aywaldron (approved):
Thanks for the docs!

slaigai (approved):

PR #550, opened 2022-05-04
Ers1249/ref counters

ashule01:
## Change description

> Adding counters for number of shared memory and redis references (not taking auto-pruning for redis refs into account). Updated tests to check these counters.
>This PR also includes a breaking change to the reference_create API which will now return a success bool and a list of ref ids that is the same length as the input data even if the reference_create fails for some of the references -- this change is to ensure that we don't have lingering shared mem references that we don't know exist

## Type of change

## Related issues

> https://elementary.atlassian.net/browse/ERS-1249

### Development

- [x] Lint rules pass locally
- [x] Application changes have been tested thoroughly
- [x] Automated tests covering modified code pass

### Security

- [x] Security impact of change has been considered
- [x] Code follows company security practices and guidelines

### Code review 


languages/python/atom/element.py
```diff
@@ -2955,6 +2978,8 @@ def _reference_create_redis(
         if not all(response):
             raise ValueError(f"Failed to create reference! response {response}")
 
+        self.counter_update(AtomRefCounters.NUM_REDIS.value, len(data))
+
```
aywaldron, line 2982: Would it be safer to take the len of `ref_ids` here in case a ref create failed on one of the data?

languages/python/atom/element.py
```diff
@@ -3036,7 +3062,7 @@ def _reference_create_shared_mem(
                 except Exception as e:
                     failed_allocs.append(str(e))
                     continue
-
+                total_data_size += datum.size
```
aywaldron, line 3065: I think this give us the number of elements in the array, not the total size in memory.

languages/python/atom/element.py
```diff
@@ -3059,6 +3085,9 @@ def _reference_create_shared_mem(
                 f"Failed to create shared mem reference! Errs: {failed_allocs}"
             )
 
+        self.counter_update(AtomRefCounters.NUM_SHARED_MEM.value, len(data))
```
aywaldron, line 3088: Same question here as above, I think taking the length of `ref_ids` is more accurate.

languages/python/atom/element.py
```diff
@@ -3461,14 +3490,23 @@ def _reference_get_shared_mem(
...
+                        # return_data is a tuple of:
+                        # data, size, width, height, depth
+                        get_return_data = atomshmem.atom_shmem_get(handle)
+
+                        # need to check the shape of the return data because
+                        # if it is < 3D then the reshape command will throw
+                        # a warning for the dimensions that are "0" in the
+                        # system
+                        shape_tuple = ()
+                        for dim in get_return_data[2:5]:
```
aywaldron, line 3502: To make this more readable could we do something like
```
data, shape, width, height, depth = atomshmem.atom_shmem_get(handle)

for dim in (width, height, depth):
    ...
```

languages/python/atom/element.py
```diff
@@ -3572,6 +3610,22 @@ def _redis_key_delete(self, *keys) -> tuple[bool, list[str]]:
...
+
+        Args:
+            keys: References to delete
+
+        Returns:
+            success (bool), list of failed keys (list of reference)
+        """
+        success, failed = self._redis_key_delete(*keys)
+
+        self.counter_update(AtomRefCounters.NUM_REDIS.value, len(failed) - len(keys))
```
aywaldron, line 3625: The idea is that `len(failed) - len(keys)`  gives us a negative number, right?

languages/python/tests/test_atom.py
```diff
@@ -1593,6 +1600,51 @@ def test_reference_multiple_shared_mem_and_redis(self, caller):
...
+        for i in range(len(all_data)):
+            assert np.array_equal(ref_data[i], all_data[i])
+
+        success, failed = caller.reference_delete(*ref_ids)
+        assert success == True
+        assert len(failed) == 0
+
+        for ref_counter in AtomRefCounters:
+            assert caller.counter_get(ref_counter.value) == 0
+
```
aywaldron, line 1647: This test looks really great :tada: 
Would it be worth testing that a second element interacts smoothly with the same counters?

aywaldron (changes requested):
Looks really good Allison, couple small ish comments. I see that we're not initializing the counters anywhere, maybe that's not required though?

languages/python/atom/element.py
ashule01: Yup! I'll add a comment to make that more explicit

languages/python/tests/test_atom.py
ashule01: Oh yeah good call on that

languages/python/atom/element.py
ashule01: Because each value in the  array is a uint8, this gives us the number of bytes

languages/python/atom/element.py
ashule01: In this case, it doesn't really matter because an exception gets thrown if there is an issue with any of the creates. len(ref_ids) is equivalent to len(data) because ref_ids are set even before the create is executed. I agree though that it is cleaner, so I'll update

ashule01: Actually, this counter should be moved above the Error that is thrown and should be incremented based on the number of responses that were successful so the counter reflects the case where some successfully insert even when an Error gets thrown

languages/python/atom/element.py
aywaldron: Ah ok perfect. Could we add a comment clarifying that?

aywaldron:
> That does mean that these counters could linger across inspections though

I think this is what we want as long as it correctly reflects the references that still exist -- we'll need to make sure that when we have elements cleaning up tags on start up, we decrement the counters appropriately. Not in the scope of this PR but it might require shared mem API updates.

languages/python/tests/test_atom.py
```diff
@@ -144,6 +148,9 @@ def responder(self, client, check_redis_end, metrics):
         yield responder, responder_name
         pytest.responder_incrementor += 1
 
+        for ref_counter in AtomRefCounters:
+            responder.counter_delete(ref_counter.value)
```
aywaldron, line 152: Do we clean these up here just to get a clean slate for testing?

languages/python/tests/test_atom.py
```diff
@@ -1426,31 +1433,31 @@ def test_parameter_list_pattern_matching(self, caller):
     def test_reference_basic(self, caller):
         caller, caller_name = caller
         data = b"hello, world!"
-        ref_id = caller.reference_create(data)[0]
-        ref_data = caller.reference_get(ref_id)[0]
+        success, ref_id = caller.reference_create(data)
+        ref_data = caller.reference_get(*ref_id)[0]
```
aywaldron, line 1437: Just curious is it necessary to call with `*ref_ids` or is it possible to pass in just `ref_ids` without the star?

aywaldron (approved):
Functionality all looks good to me :+1: couple small questions/suggestions on comments. Thanks for working through the API changes!

languages/python/tests/test_atom.py
```diff
@@ -3224,15 +3331,14 @@ def test_fail_to_add_ref_to_destroyed_shared_mem(self, caller):
...
-            ret = caller.reference_create(data, ref_type=AtomRefType.SHARED_MEM)[0]
-        except Exception as e:
-            ret = str(e)
-
-        assert "ATOM_SHMEM_DOES_NOT_EXIST" in ret
+        success, ref_ids = caller.reference_create(
+            data, ref_type=AtomRefType.SHARED_MEM
+        )
+        assert success == False
+        assert np.array_equal(ref_ids, [None])
```
aywaldron, line 3341: We could also check here that the counter hasn't incremented.

languages/python/tests/test_atom.py
ashule01: Yeah because any test that makes a reference will create this counter so if we don't clean them up the lingering ones throw an assertion error in check_end_redis

languages/python/tests/test_atom.py
ashule01: so you can call it with ref_id only if it is a single ref id as a string or you can call it with *ref_ids where ref_ids is a list. The change here is because we used to do reference_create()[0] which would make it so ref_id was a string, but since we are now returning a tuple (bool, list of ids) we can't use the [0] on the same line to get the first value anymore. This code would work in one of two ways:
```
success, ref_id = caller.reference_create(data) #ref_id here is a list
ref_data = caller.reference_get(*ref_id)[0]
```

```
success, ref_ids = caller.reference_create(data) #ref_ids here is a list
ref_id = ref_ids
ref_data = caller.reference_get(ref_id)[0]
```

aywaldron: Ok thanks for explaining!

languages/python/atom/element.py
```diff
@@ -3033,7 +3039,7 @@ def _reference_create_shared_mem(
             # Serialize
             failed_allocs = []
             for n, datum in enumerate(data):
-                ref_ids.append(None)
+                # ref_ids.append(None)
```
aywaldron, line 3042: Can probably delete this

languages/python/atom/element.py
```diff
@@ -3461,14 +3514,25 @@ def _reference_get_shared_mem(
...
                         )
 
-                        # setup image
+                        # need to check the shape of the return data because
+                        # if it is < 3D then the reshape command will throw
+                        # a warning for the dimensions that are "0" in the
+                        # system
+                        shape_tuple = ()
+                        for dim in (width, height, depth):
+                            if dim != 0:
```
dpipemazo, line 3529: this is an interesting edge/corner case. Do we have unit tests covering how we handle arrays with zero-size dimensions? Might be good to add it in if not, since this seems like it can subtly break.

dpipemazo (approved):
Looks great -- just some minor nits/questions!

languages/python/atom/element.py
ashule01: Do you mean where one or more of the dimensions is 0 because we have the following for 1, 2, 3, and 4D data (in case of the 4D we get a failure):
1D: https://github.com/elementary-robotics/atom/blob/3d08c1cbb2a115ea3fd8178de18ba536fedd4ebb/languages/python/tests/test_atom.py#L1443
2D: https://github.com/elementary-robotics/atom/blob/3d08c1cbb2a115ea3fd8178de18ba536fedd4ebb/languages/python/tests/test_atom.py#L1489
3D: https://github.com/elementary-robotics/atom/blob/3d08c1cbb2a115ea3fd8178de18ba536fedd4ebb/languages/python/tests/test_atom.py#L1499
4D: https://github.com/elementary-robotics/atom/blob/3d08c1cbb2a115ea3fd8178de18ba536fedd4ebb/languages/python/tests/test_atom.py#L1509

dpipemazo: Nice! I had no particular thoughts in mind; just thought it was the kind of clause in code we should ensure we test around the corners of. If you're confident we have the tests we're in good shape. Thanks!

PR #552, opened 2022-05-12
atom: basic capabilities implementation as ABC

dpipemazo:
First commit to get something working on capabilities. Testing
out new structure on abstract base class to clean up/improve element
code and looking for feedback.

## Change description

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Security

### Code review 


languages/python/atom/element.py
```diff
@@ -153,20 +126,6 @@ class ElementConnectionTimeoutError(redis.exceptions.TimeoutError):
...
-            self.message = None
-
-    def __str__(self):
-        if self.message:
-            return f"Atom Error: {self.message}"
-        else:
-            return "An Atom Error Occurred"
-
-
 class SetEmptyError(Exception):
```
jgauth, line 129: could this be moved to exceptions.py?

languages/python/atom/element.py
```diff
@@ -248,6 +208,16 @@ def __init__(
                 out while waiting for data back over a Redis connection.
```
jgauth, line 208: capabilities parameter not in docstring

jgauth (commented):
overall, ABC usage here seems to fit very well

languages/python/atom/element.py
dpipemazo: We could move this. There's a few things to think through:
1. Is moving it the cleaner/better logical place for it? I'd agree with you and say yes.
2. Do we risk introducing additional risk/errors into our codebases by moving it? I worry about this a bit -- if we have code that imports `SetEmptyError` to catch it, it'll throw if we change the location and don't change the code. I _think_ if anything doing this is running `pyright` then it should get caught in CI, but we risk having code break silently in prod.

What do you think?

jgauth: I _think_ that would depend on how the other codebases are importing the class? E.g. `from atom.element import SetEmptyError` would fail if we move it to exceptions.py. However we do have it in the [package init](https://github.com/elementary-robotics/atom/blame/latest/languages/python/atom/__init__.py#L2) so the other codebases could be doing `from atom import SetEmptyError` which would be fine (assuming we update `__init__.py`). Does that seem right?

dpipemazo: Ah nice, I didn't realize we had it handled in `__init__.py`! 

I also audited our codebases using Github search and it seems there's good news that no repos outside of Atom are using this anyway :smile: . 

I'll go ahead and move it. Thanks!

languages/python/atom/element.py
```diff
@@ -245,8 +187,20 @@ def __init__(
...
 
+        #
+        # Properties
+        #
+        #   These are needed for the scheme where we implement subcomponents
+        #   of Atom as abstract base classes. The ABCs will need to use some
+        #   fields on the element, and implementing these fields of as
+        #   properties gives the ABC some guarantees that the fields are
+        #   implemented and of a certain type
+        #
```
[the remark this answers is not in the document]
dpipemazo: Ah, yes. thank you for catching this! Took me a few tries to get the ABC properties set up properly and I missed cleaning this up.

aywaldron (approved):
Thanks for adding the docs Dan!

ashule01 (approved):
Thanks for all the comments and documentation Dan! Very easy to follow :+1:

PR #554, opened 2022-05-13
Ers2843/remove uint8

ashule01:
## Change description

> Remove value check for uint8 type to checking the dtype of the np array instead. Added tests to compare the timing for redis and shared mem references.

## Type of change
- [x] Bug fix (fixes an issue)

## Related issues

> https://elementary.atlassian.net/browse/ERS-2843

### Development

- [ ] Lint rules pass locally

### Security

### Code review 


dpipemazo (approved):

languages/python/tests/test_atom.py
```diff
@@ -2118,6 +2118,79 @@ def test_reference_ignore_serialization(self, caller):
...
+        redis_create = after_redis_create - before_redis_create
+        shm_get = after_shm_get - after_shm_create
+        redis_get = after_redis_get - after_redis_create
+        shm_delete = after_shm_delete - after_shm_get
+        redis_delete = after_redis_delete - after_redis_get
+
+        # shm is within 10% of the time it takes for redis
+        # to complete reference create, get, and delete
+        assert (
+            shm_create <= 1.1 * redis_create
```
aywaldron, line 2187: Maybe this check should be something like this instead:
```suggestion
            shm_create - redis_create <= 1.1 * min(shm_create, redis_create)
```
so we are checking that the difference between the times is within 10% of which ever is smaller? Since as written, if the redis time became very large by mistake, the current check would still pass and we'd miss the regression.

aywaldron: Not sure if 10% is the right number, maybe its not and maybe we should also be checking a hard threshold in case they _both_ increased drastically.

dpipemazo: I think we should just have a hard threshold here, maybe 100ms, around a fairly large piece of data. Making it relative will be asking for trouble/flaky tests, and we really care more about not adding some undue overhead.

ashule01: Thanks for the feedback on that Dan. Do we care about the timing when running tests locally vs circle because the times are looking pretty dramatically different

languages/python/atom/element.py
```diff
@@ -3527,24 +3528,31 @@ def _reference_get_shared_mem(
...
+                        if dim != 0:
+                            shape_tuple = shape_tuple + (dim,)
+                        else:
+                            break
+
+                    # reshape the data, but can't use "reshape" function
+                    # because that copies the data
+                    try:
+                        data.shape = shape_tuple
+                    except:
```
dpipemazo: It would be ideal to get into the habit of catching the specific exception thrown. It's generally most Pythonic to let something we're not trying to explicitly handle still throw us an exception, and this bare try/except can cause us to squash some errors that we're not expecting.

languages/python/atom/element.py
```diff
@@ -3530,24 +3527,39 @@ def _reference_get_shared_mem(
...
+                    except ValueError as e:
+                        self.logger.warning(
+                            f"Cannot reshape data as the total size of the \
+                                array must be unchanged: {str(e)}"
+                        )
+                    except AttributeError as e:
+                        self.logger.warning(
+                            f"Cannot perform inplace modification of data \
+                                because of incompatible shape: {str(e)}"
+                        )
```
aywaldron, line 3560: So the behavior here in case the reshape fails is to return the data in its original shape, is that right? Could there be any issues then when we try to use it, or do we think generally it will be ok? Mainly wondering if its better to return None when we can't do the reshape, I'm not sure

languages/python/tests/test_atom.py
```diff
@@ -2118,6 +2117,42 @@ def test_reference_ignore_serialization(self, caller):
...
+        after_delete = time.time_ns() / 1e6
+
+        time_create = after_create - before_create
+        time_get = after_get - after_create
+        time_delete = after_delete - after_get
+
+        # These are very generous time bounds because we don't want these
+        # tests to be flaky (in ms)
+        time_tuple = (25, 10, 5) if ref_type == AtomRefType.SHARED_MEM else (50, 100, 5)
+        create_compare, get_compare, delete_compare = time_tuple
```
aywaldron, line 2147: Might be more clear to call these "thresholds" or something like that

languages/python/atom/element.py
dpipemazo: If we can't reshape I think something went wrong in our internal implementation of shared memory -- it shouldn't happen and should likely raise?

aywaldron: When a GET fails for some reason we return None for that key instead of raising since we may be getting multiple references per call. But yeah I'm also thinking we should probably indicate it failed instead of just returning the raw data.

dpipemazo: That's slightly different as there's an external dependency on Redis which can fail. We have an internal dependency on our own code here which should be rock solid, so it makes more sense to raise/yell loudly if this happens since it really shouldn't by design.

aywaldron: So if a user tries to get multiple references in a single call and one datum fails to reshape they should not receive the data for the other keys? I think its preferable that they still get the good data to continue working with, since there's no reason to fail on all keys if just one is causing an issue.

dpipemazo: I think we're maybe not on the same page about the root cause of the reshape failing -- to me this can't happen unless something is very wrong in our implementation. Once we've written/tested the code there's really no reason this error should happen, so my preference is to have this sort of thing yell loudly and fail since something is very wrong.

dpipemazo: If it's an error that can happen in production due to things out of our control (redis being down, etc.) I would definitely agree with your strategy, but this seems like a core component of Atom that just shouldn't break.

aywaldron: I don't know if its possible in production but think its better to plan for the worst in case it is. Our tests would catch it failing regardless of how we let it fail.

dpipemazo: I think we may just disagree here which is fine; I'm looking for the implementation to guarantee this can't happen which should be possible. I don't really agree that we should be error-handling around things which would be bugs in our own code. 

We can take it however you'd like and it's not a huge deal, just wanted to explain the philosophy.

aywaldron: That makes sense, I'm also fine with either way. I guess I don't feel familiar enough with shared memory or Iouri's library to be confident in what fail cases are or are not possible to unexpectedly hit in the field but if we're very certain this one is impossible then I feel ok with letting it raise.

dpipemazo: I guess I see it as:
- We have a numpy array of a certain shape on the way in
- We store the shape and the array in shared mem
- We get the array back from shared mem and apply the stored shape

Not sure there are a lot of paths where it could/should fail

aywaldron: Yeah, you're right it should always work since even if the shape in the header was corrupted or something, its used to get the data so should always match the shape that's returned.

dpipemazo: I'm not sure I follow the last part, but also I think it's preferable where, in the event the header is corrupted, this code raises rather than prints a warning. To me, there's something very wrong and our whole app is in a lot of trouble if shared memory is getting corrupted.

languages/python/tests/test_atom.py
ashule01: Updated!

aywaldron (approved):
Thanks Allison!

PR #565, opened 2022-06-17
Remove autoprune flag

ashule01:
## Change description

> The flag was used to prevent the AtomRefCounter from incrementing, so we switched to check if a max len was set, if yes then do not increment counter

## Type of change
- [x] Bug fix (fixes an issue)

## Related issues

> https://elementary.atlassian.net/browse/ERS-3130

### Development

- [ ] Lint rules pass locally

### Security

### Code review 


languages/python/atom/element.py
```diff
@@ -3162,13 +3165,13 @@ def _reference_create_shared_mem(
...
-        if autoprune is False:
-            self.counter_update(
-                AtomRefCounters.NUM_SHARED_MEM.value, len(ref_ids) - len(failed_allocs)
-            )
+            # we will not be using the ref counters for autoprune
+            # because we will assume they will automatically get cleaned
+            # up by maxlen, so only increment counter when no max_len is set
+            if counter_increment > 0:
+                self.counter_update(
+                    AtomRefCounters.NUM_SHARED_MEM.value, counter_increment
```
aywaldron, line 3173: Was there a reason the `len(ref_ids) - len(failed_allocs)` no longer worked?

ashule01: Because we do the max_len calculation based on if any max_len had been defined now or previously, there could be a combination of references where they don't have a max_len defined on this specific call but had defined one previously, so we need to track the number of non-autopruned tags separately from the ref_ids-failed calculation

aywaldron: Ok; I'm not sure I follow completely because every time this method it gets a single tag with a single `max_len` so the calculated `tag_max_len` should be the same for all refs in the loop -- but the implementation you have here also seems fine :+1:

ashule01: Ah true true you're right :+1: I'll update that

languages/python/atom/element.py
```diff
@@ -3162,13 +3159,14 @@ def _reference_create_shared_mem(
...
-                AtomRefCounters.NUM_SHARED_MEM.value, len(ref_ids) - len(failed_allocs)
-            )
+            # we will not be using the ref counters for autoprune
+            # because we will assume they will automatically get cleaned
+            # up by maxlen, so only increment counter when no max_len is set
+            if tag_max_len == 0:
+                self.counter_update(
+                    AtomRefCounters.NUM_SHARED_MEM.value,
+                    len(ref_ids) - len(failed_allocs),
+                )
```
aywaldron, line 3169: Do we want this outside of the metrics pipeline?

